PR #3, opened 2021-11-25
Use rdseed/rdrand to generate random

yang8621:


syan10 (commented):
suggest to add the rand ralated codes into a slipt file, e.g. rand.c

syan10 (approved):
LGTM

PR #8, opened 2021-12-02
add a template implementation for NAPI_fucnction

syan10:
NAPI_CreateKey
NAPI_Encrypt
NAPI_Decrypt

test: passed unittest
Signed-off-by: Yan, Shaopu <shaopu.yan@intel.com>

### DIFF
--- a/core/App/ehsm_napi.cpp
+++ b/core/App/ehsm_napi.cpp
@@ -1,0 +1,312 @@
+/*
+ * Copyright (C) 2021-2022 Intel Corporation
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   1. Redistributions of source code must retain the above copyright
+ *      notice, this list of conditions and the following disclaimer.
+ *   2. Redistributions in binary form must reproduce the above copyright
+ *      notice, this list of conditions and the following disclaimer in
+ *      the documentation and/or other materials provided with the
+ *      distribution.
+ *   3. Neither the name of Intel Corporation nor the names of its
+ *      contributors may be used to endorse or promote products derived
+ *      from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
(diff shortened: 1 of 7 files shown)


core/App/ehsm_napi.cpp
```diff
@@ -0,0 +1,317 @@
...
+#include "ehsm_napi.h"
+
+#include "serialize.h"
+
+//#include "ehsm_provider.h"
+
+using namespace std;
+using namespace EHsmProvider;
+
+static char* StringTochar(std::string str)
```
yang8621, line 42: StringToChar.  
Why not use str.c_str() directly?

core/App/ehsm_napi.cpp
```diff
@@ -0,0 +1,317 @@
...
+        retChar = (char *)malloc(len * sizeof(uint8_t));
+        if(retChar != nullptr){
+            memset(retChar, 0, len);
+            memcpy(retChar, str.c_str(), len);
+        }
+    }
+    return retChar;
+}
+
+std::string intToString(int value)
```
yang8621, line 56: use existing std::to_string()

core/App/ehsm_napi.cpp
syan10: str.c_str() return a const char*, not match with the function return type (char*)

core/App/ehsm_napi.cpp
```diff
@@ -0,0 +1,317 @@
...
+{
+    char buf[32] = {0};
+    snprintf(buf, sizeof(buf), "%u", value);
+    std::string str = buf;
+    return str;
+}
+
+struct RetJsonObj
```
yang8621, line 64: Do we really need to format the return value as JSON in NAPI layer? @syan10 
JS layer should format the JSON obj according to the different errors like checking the signature, time, API validation, etc.

syan10: but how to return the error code and such above infomations? since currently the neusoft guys said it only can work with the return char*, however, we need more extra infomations need to return. so a json struct string should be ok.

core/App/ehsm_napi.cpp
yang8621: so use str.copy()

yang8621: OK.   Change StringTochar -> StringToChar

core/App/ehsm_napi.cpp
yang8621: So why line@156,157 is commented out for successful return?

syan10: just make the unit test case happy, because it need to add c++ json parser, otherwise the test cases can't get the correct prameters.

after the neusoft guys added the c++ json parser, it don't need to be commented out and then remove the StringtoChar() function.

core/App/ehsm_napi.cpp
```diff
@@ -0,0 +1,320 @@
...
+        if(jsonStr.size() > 0){
+            jsonStr += ",";
+        }
+        jsonStr += "\""+key+"\" : " + "\""+data+"\"";
+    };
+
+    char* toChar()
+    {
+        std::string retString = "{";
+        retString += "\"code\":" + intToString(code);
```
yang8621, line 83: retString += "\"code\":" + std::to_string(code);

yang8621 (approved):

PR #20, opened 2021-12-08
Add a simple json parser, to make it work in the backend c++ unittest.

Liang1XZhao:
ehsm_napi.cpp add parseStringData function .

### DIFF
--- a/core/App/ehsm_napi.cpp
+++ b/core/App/ehsm_napi.cpp
@@ -80,6 +80,22 @@ typedef struct {
         retString += "}";
 		return StringToChar(retString);
 	};
+
+    char* parseStringData(char* jsonChar, std::string key) const{
+        std::string retVal;
+        std::string jsonString = jsonChar;
+        key = "\"" + key + "\"";
+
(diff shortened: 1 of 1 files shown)


core/App/ehsm_napi.cpp
```diff
@@ -80,6 +80,22 @@ typedef struct {
         retString += "}";
 		return StringToChar(retString);
 	};
+
+    char* parseStringData(char* jsonChar, std::string key) const{
```
syan10, line 84: the function name not very fitable, for the implementation , it seems just for using to retrieve the "key" value, right?


and who will call it?

Liang1XZhao: yes. It simply parses string data, which will be used in unit tests.
You can see it in ehsm_core_test.cpp of the next PR.

syan10: just see that in the json struct, there's already has a function named as "void addData(string key, string data)", so why not add another function as "void readData(string key, string data)"

Also, you need to re-submit the PR together with the changes in the unittest code.

Liang1XZhao:
resubmit after repair

PR #21, opened 2021-12-08
dockerfile node service

Liang1XZhao:
dockerfile add node service

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -51,18 +51,35 @@ RUN apt-get update &&apt-get install -y \
     libsgx-ra-network \
     libsgx-ra-uefi
 
-# The prebuild app
-WORKDIR /home/ehsm/ehsm-core
-COPY bin/ehsm-core/ehsm-core /home/ehsm/ehsm-core/ehsm-core
-COPY bin/ehsm-core/libenclave-ehsm-core.signed.so /home/ehsm/ehsm-core/libenclave-ehsm-core.signed.so
 
 # Build App from source
-#RUN git clone https://github.com/intel/ehsm.git \
-#    && cd ehsm \
-#    && make
+WORKDIR /home
+RUN git clone https://github.com/intel/ehsm.git \
+   && cd ehsm \
+   && make
+
+# Install node & npm packages
+RUN wget https://nodejs.org/dist/v16.13.0/node-v16.13.0-linux-x64.tar.xz \
+    && tar xf  node-v16.13.0-linux-x64.tar.xz \
+    && mv node-v16.13.0-linux-x64/ /usr/local/nodejs \
+    && ln -s /usr/local/nodejs/bin/node /usr/local/bin \
+    && ln -s /usr/local/nodejs/bin/npm /usr/local/bin
+
+
+# Install ehsm-service dependence packages
+WORKDIR /home/ehsm/ehsm_kms_service
+RUN cp /home/ehsm/bin/ehsm-core/libehsmnapi.so . \
+    && cp /home/ehsm/bin/ehsm-core/libenclave-ehsm-core.signed.so .
+RUN npm install
+
+# image port
+EXPOSE 3000
+
+# run ehsm_kms_service
+CMD ["node","ehsm_kms_server.js"]
 
 
 #RUN adduser -q --disabled-password --gecos "" --no-create-home sgxuser
 #USER sgxuser
 
-CMD tail -f /dev/null
+# CMD tail -f /dev/null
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -51,18 +51,33 @@ RUN apt-get update &&apt-get install -y \
...
+RUN wget https://nodejs.org/dist/v16.13.0/node-v16.13.0-linux-x64.tar.xz \
+    && tar xf  node-v16.13.0-linux-x64.tar.xz \
+    && mv node-v16.13.0-linux-x64/ /usr/local/nodejs \
+    && ln -s /usr/local/nodejs/bin/node /usr/local/bin \
+    && ln -s /usr/local/nodejs/bin/npm /usr/local/bin
+
+
+# Install ehsm-service dependence packages
+WORKDIR /opt/intel/ehsm/ehsm_kms_service
+RUN cp /opt/intel/ehsm/bin/ehsm-core/*so .
```
syan10, line 70: "RUN cp /opt/intel/ehsm/bin/ehsm-core/*so ."

don't copy all the *.so, just copy the necessary file(libehsmnapi.so, libenclave-ehsm-core.signed.so)

docker/Dockerfile
```diff
@@ -51,18 +51,33 @@ RUN apt-get update &&apt-get install -y \
...
+RUN npm install
```
syan10, line 71: i'm not sure what pkgs that cmd "npm install" will try to install?

can we only install the necessary files?

docker/build_and_run.sh
```diff
@@ -4,5 +4,5 @@ docker build --build-arg https_proxy=$https_proxy \
 
 # Another container should expose AESM and its socket in aesmd-socket volume.
 # Replace /dev/sgx/enclave with /dev/isgx if you use the Legacy Launch Control driver
-docker run --env http_proxy --env https_proxyi --device=/dev/sgx/enclave -v aesmd-socket:/var/run/aesmd -it ehsmcore bash
+docker run --env http_proxy --env https_proxyi --device=/dev/sgx/enclave -v aesmd-socket:/var/run/aesmd -it -P ehsmcore
```
syan10, line 7: --env https_proxyi is a typo.

correct it as "--env https_proxy"

syan10: what's the "-P ehsmcore" means?

docker/Dockerfile
```diff
@@ -51,18 +51,33 @@ RUN apt-get update &&apt-get install -y \
...
-# The prebuild app
-WORKDIR /home/ehsm/ehsm-core
-COPY bin/ehsm-core/ehsm-core /home/ehsm/ehsm-core/ehsm-core
-COPY bin/ehsm-core/libenclave-ehsm-core.signed.so /home/ehsm/ehsm-core/libenclave-ehsm-core.signed.so
 
 # Build App from source
-#RUN git clone https://github.com/intel/ehsm.git \
-#    && cd ehsm \
-#    && make
+RUN git clone https://github.com/intel/ehsm.git \
```
syan10, line 56: why not set a default workdir?

docker/Dockerfile
Liang1XZhao: done

docker/Dockerfile
Liang1XZhao: npm install downloads the required dependencies according to package.json

docker/Dockerfile
Liang1XZhao: it will be fix

docker/build_and_run.sh
Liang1XZhao: -P 
Container internal ports are randomly mapped to host ports

docker/Dockerfile
Liang1XZhao: it will be fix

syan10 (changes requested):
no comments for the changes now.

But, the commit msg tile need to be updated, please note that the title should be a short sentence.
"dockerfile node service" what does this mean? you can adjust it as
"add node service support in the dockerfile"

ehsm_kms_service/ehsm_kms_server.js
```diff
@@ -258,5 +258,7 @@ app.post('/ehsm', function (req, res) {
   }
 })
 
-app.listen(port, () => {})
+app.listen(port, () => {
+  console.log(`ehsm_ksm_service application listening at http://localhost:${port}`)
```
syan10, line 262: change to
console.log(`ehsm_ksm_service application listening at ${port}...`)

and if there's some one connect to it,  it's better to print out the target ip.

Liang1XZhao: it will be fix

Liang1XZhao:
Resubmit after repair

PR #22, opened 2021-12-08
NAPI_Initialize NAPI_Finalize test_AES128

Liang1XZhao:
c++ complete NAPI_Initialize 、NAPI_Finalize、 test_AES128
node-service add NAPI_Initialize、NAPI_Finalize

### DIFF
--- a/core/App/ehsm_core_test.cpp
+++ b/core/App/ehsm_core_test.cpp
@@ -42,577 +42,60 @@
 
 using namespace EHsmProvider;
 
-static void dump_data(uint8_t *data, uint32_t datalen) {
-    uint32_t i;
-    std::string encode_str;
-
-    printf("datalen=%d, data is:\n", datalen);
-
-    encode_str = base64_encode(data, datalen);
-    printf("%s\n", encode_str.c_str());
-
-#if 0
-    std::string decode_str;
-    decode_str = base64_decode(encode_str);
-    printf("decode_str.size=%ld, data is:\n", decode_str.size());
-
-    for (i=1; i<=decode_str.size(); i++) {
-        printf("%d\t", decode_str.data()[i-1]);
-        if (i%16 == 0)
-            printf("\n");
-    }
-
-    std::string encode_str2;
-    encode_str2 = base64_encode((uint8_t*)decode_str.data(), decode_str.size());
-    printf("str=%s\n", encode_str2.c_str());
-#endif
-}
-
-ehsm_status_t testAES()
-{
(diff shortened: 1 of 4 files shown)


core/App/ehsm_napi.cpp
```diff
@@ -36,53 +36,37 @@
...
-        retString += "}";
-		return StringToChar(retString);
-	};
-} RetJsonObj;
 
-extern "C" {
+/*
+destory the enclave
+*/
+void NAPI_Finalize(){
```
yang8621, line 67: not found the correct place to call?

Liang1XZhao: This method is called in nodejs.

yang8621: I only find ehsm_napi.NAPI_Initialize() is called.   Where is NAPI_Finalize()?

Liang1XZhao: It will be fix.

core/App/ehsm_core_test.cpp
```diff
@@ -42,577 +42,60 @@
...
-    SAFE_FREE(plaintext.data);
-    SAFE_FREE(ciphertext.data);
-    SAFE_FREE(plaintext2.data);
-
-    printf("============testAES Done==========\n");
-
-    return ret;
-}
-
-ehsm_status_t testRSA()
```
syan10, line 169: don't delete the original test cases unless you have already implemented the new one.

core/App/ehsm_core_test.cpp
```diff
@@ -42,577 +42,60 @@
...
-    SAFE_FREE(cipher_datakey.data);
-    SAFE_FREE(cipher_datakey2.data);
-    SAFE_FREE(plaintext1.data);
-    SAFE_FREE(plaintext2.data);
-
-    printf("============testGenerateDataKey done==========\n");
-    return ret;
+    
+out:
+    printf("============test_AES128 end==========\n");
```
syan10, line 96: seems that you didn't free the memory!！

the NAPI_xxx function will return a json string, i'm not sure whether the nodejs side will free them automatically? and for the c++ unittest, it seems that you need to free them by yourself.

syan10 (changes requested):
-1

core/App/ehsm_core_test.cpp
Liang1XZhao: It will be fix

Liang1XZhao:
resubmit after repair

PR #25, opened 2021-12-09
add node service support in the dockerfile

Liang1XZhao:
- Dockerfile new node environment
 - Service start print target IP and port

### DIFF
--- a/ehsm_kms_service/ehsm_kms_server.js
+++ b/ehsm_kms_service/ehsm_kms_server.js
@@ -258,5 +258,20 @@ app.post('/ehsm', function (req, res) {
   }
 })
 
-app.listen(port, () => {})
+const  getIPAdress = () => {
+  var interfaces = require('os').networkInterfaces();　　
+  for (var devName in interfaces) {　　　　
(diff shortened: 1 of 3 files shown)


ehsm_kms_service/ehsm_kms_server.js
```diff
@@ -258,5 +258,20 @@ app.post('/ehsm', function (req, res) {
...
+          var alias = iface[i];
+          if (alias.family === 'IPv4' && alias.address !== '127.0.0.1' && !alias.internal) {
+              return alias.address;
+          }
+      }　　
+  }
+}
+
+app.listen(port, () => {
+  console.log(`ehsm_ksm_service application listening at ${getIPAdress()}:${port}`)
```
syan10, line 275: you need to print out the target ip that who connected to the server, not the server's IP

285241: This is the log printed once when the server is started, which is used to display the IP address of the currently running nodejs server。

syan10 (approved):

yang8621 (approved):

PR #26, opened 2021-12-09
complete NAPI_Initialize NAPI_Finalize test_AES128

Liang1XZhao:
change function name parseStringData -> readData
in the ehsm_kms_service initialize and finalize
test_kms_with_rest.py : plaintext_base64 -> plaintext

### DIFF
--- a/core/App/ehsm_core_test.cpp
+++ b/core/App/ehsm_core_test.cpp
@@ -68,103 +68,64 @@ static void dump_data(uint8_t *data, uint32_t datalen) {
 #endif
 }
 
-ehsm_status_t testAES()
+/*
+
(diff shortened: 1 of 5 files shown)


syan10 (approved):
LGTM

syan10 (commented):
combine the 2 commits due to they are doing one thing and the changes seems not too much, and also need to change the commit title as
“
add  NAPI_Initialize and NAPI_Finalize  support for  nodejs

nodejs server will call the initialize to create enclave when it
first start and call the finalize to destory the enclave while the
service is down.

- add a new unit test based on the NAPI_xxx call.

test:  passed the  unittest
”

PR #29, opened 2021-12-10
add NAPI_GenerateDataKeyWithoutPlaintext support for nodejs

Liang1XZhao:
nodejs server call NAPI_GenerateDataKeyWithoutPlaintext

add a new unit test test_generate_datakey
- NAPI_GenerateDataKey call
- NAPI_GenerateDataKeyWithoutPlaintext call

test: passed the unittest

### DIFF
--- a/core/App/ehsm_napi.cpp
+++ b/core/App/ehsm_napi.cpp
@@ -311,8 +311,7 @@ char* NAPI_GenerateDataKey(const char* cmk_base64,
         const uint32_t keylen,
         const char* aad)
 {
-    string decode_str;
-    string decode_cipher;
+    string cmk_str;
     ehsm_status_t ret = EH_OK;
     RetJsonObj retJsonObj;
     ehsm_keyblob_t masterkey;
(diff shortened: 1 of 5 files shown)


core/App/ehsm_napi.cpp
```diff
@@ -375,26 +374,97 @@ char* NAPI_GenerateDataKey(const char* cmk_base64,
...
+        retJsonObj.addData("plaintext_base64", plaintext_base64); 
+		if(ciphertext_base64.size() > 0){
+			retJsonObj.addData("ciphertext_base64", ciphertext_base64);
+			SAFE_FREE(masterkey.keyblob);
+			SAFE_FREE(plaint_datakey.data);
+			SAFE_FREE(cipher_datakey.data);
+			return retJsonObj.toChar();
+		}
+    }
+
```
syan10, line 386: this code can be refined as this below:
"
if((plaintext_base64.size() > 0)  && (ciphertext_base64.size() > 0) ){
        retJsonObj.addData("plaintext_base64", plaintext_base64); 
        retJsonObj.addData("ciphertext_base64", ciphertext_base64);
    }
"

and same with the other functions,  please remove the duplicate codes when success return.
"
SAFE_FREE(xxx);
SAFE_FREE(xxx);
return retJsonObj.toChar();

syan10 (changes requested):
the commit title need to be refined(this API not similar with initialize/finalize which just for nodejs call), however, this API will exposed to public as a restful API, so you can reference the below msg for other rest API commits:
"
add support for the GenerateDataKeyWithoutPlaintext restful API

this restful API will exposed to public as:
POST <ehsm_srv_address>/ehsm?Action=GenerateDataKeyWithoutPlaintext 

test: passed the unittest
"

PR #33, opened 2021-12-10
add support for the Sign and Verify restful API

Liang1XZhao:
this restful API will exposed to public as:
- POST <ehsm_srv_address>/ehsm?Action=Sign
- POST <ehsm_srv_address>/ehsm?Action=Verify
test: passed the unittest

### DIFF
--- a/core/App/ehsm_napi.h
+++ b/core/App/ehsm_napi.h
@@ -53,11 +53,12 @@ static char* StringToChar(string str)
 typedef struct {
     const int CODE_SUCCESS = 200;
     const int CODE_FAILED = 500;
+
     int code = CODE_SUCCESS;
     std::string msg = "success!";
     std::string jsonStr;
 	
-	void setCode(int code){code = code;};
+	void setCode(int newCode){code = newCode;};
 	void setMessage(string message){msg = message;};
 	void addData(string key, uint32_t data) {
         if(jsonStr.size() > 0){
@@ -71,8 +72,18 @@ typedef struct {
         }
         jsonStr += "\""+key+"\":" + "\""+data+"\"";
     };
+	void addData(string key, bool data) {
+        if(jsonStr.size() > 0){
+            jsonStr += ",";
+        }
+        if(data){
+            jsonStr += "\""+key+"\":true";
+        } else {
+            jsonStr += "\""+key+"\":false";
+        }
+    };
 
-  char* toChar() {
+    char* toChar() {
         std::string retString = "{";
         retString += "\"code\":" + std::to_string(code);
         retString += ",\"message\":\"" + msg;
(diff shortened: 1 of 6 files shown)


ehsm_kms_service/test/test_kms_with_rest.py
```diff
@@ -2,157 +2,146 @@
...
+
+appid = 202112101919
+nonce = ''
+nonce = nonce.join(random.choice('0123456789') for i in range(16))
+timestamp = int(time.time())
+
+
+params = {}
+params["appid"] = appid
+params["nonce"] = nonce
```
yang8621, line 16: nonce, timestamp, and signature should 1:1 mapping with each rest call.

ehsm_kms_service/test/test_kms_with_rest.py
```diff
@@ -2,157 +2,146 @@
...
+params["timestamp"] = timestamp
+params["sign"] = 'timestamp'
```
yang8621, line 18: Hardcode a key, paired with appid, and calculate the HMAC-SHA256 as the signature.

ehsm_kms_service/ehsm_kms_params.js
```diff
@@ -72,6 +72,48 @@ const ehsm_kms_params = {
     },
     aad,
   },
+  GenerateDataKeyWithoutPlaintext: {
+    cmk_base64: {
+      type: 'string',
+      maxLength: 1024,
```
syan10, line 78: the cmk's max length is not 1024, it's decided by the keyspec, for now:
if the keyspec=AES128, then the length is 64,  if the keyspec=RSA3072, then the length seems is 2168

syan10: same issue for the above APIs define

ehsm_kms_service/ehsm_kms_server.js
```diff
@@ -53,6 +53,10 @@ const _checkSign = function (req,res,next){
    const action = req.query.Action;
   const { payload } = req.body;
   const currentPayLoad = ehsm_kms_params[action];
+  if (!currentPayLoad) {
```
syan10, line 56: indend issue

ehsm_kms_service/ehsm_kms_params.js
Liang1XZhao: it will be fix

ehsm_kms_service/test/test_kms_with_rest.py
Liang1XZhao: it will be fix

yang8621:
BTW, the encryption/decryption of RSA3072 will also be added in the next patches, right?

core/App/ehsm_napi.h
```diff
@@ -82,20 +93,41 @@ typedef struct {
...
  
-  static char* readData(char* jsonChar, std::string key){
+    static char* readData_string(char* jsonChar, std::string key){
         std::string retVal;
         std::string jsonString = jsonChar;
         key = "\"" + key + "\"";
 
-        int startIndex = jsonString.find(key) + key.size() + 1;
-        std::string subStr = jsonString.substr(startIndex);
+        int resultIndex = jsonString.find("\"result\"") + 8;
```
yang8621, line 101: What does 8 mean?  Could add some comments or use sizeof("xxxx")/strlen("xxx") instead?

ehsm_kms_service/ehsm_kms_params.js
syan10: you can just set the hardcode max length for now(e.g. 4K, don't set the same length with plaintex data), maybe later we can refine the cmk's parameter to the cmk's ID which is a fix length.

core/App/ehsm_napi.h
Liang1XZhao: It will be fix.

ehsm_kms_service/test/test_kms_with_rest.py
```diff
@@ -2,157 +2,146 @@
 import json
 import argparse
 import base64
+import time
+import random
+
+appid = 202112101919
+nonce = ''
+nonce = nonce.join(random.choice('0123456789') for i in range(16))
```
yang8621, line 10: It's ranged from 0~0x2386F26FC0FFFF, 6 bytes~7 bytes.

I checked the design doc, where the nonce length is defined as 16. Does the 16 mean 16 Bytes?
If yes, nonce = ''.join(random.choice('0123456789ABCDEF') for i in range(32))
I think 4 bytes of the length is enough: nonce = base64(os.urandom(4))

yang8621: int.from_bytes(os.urandom(4), "big")

Liang1XZhao: It will use int.from_bytes(os.urandom(16), "big")

syan10:
> BTW, the encryption/decryption of RSA3072 will also be added in the next patches, right?

Liang1XZhao:
> BTW, the encryption/decryption of RSA3072 will also be added in the next patches, right?


> > BTW, the encryption/decryption of RSA3072 will also be added in the next patches, right?

yes

PR #42, opened 2021-12-21
add the verification of the parameters

Liang1XZhao:
- check sign
 - check parameter
 - add postman
test: passed the unittest


core/App/ehsm_napi.cpp
```diff
@@ -224,41 +254,64 @@ char* NAPI_Decrypt(const char* cmk_base64,
...
+    string ciphertext_str;
+    int cmk_len;
+    int ciphertext_len;
+    int aad_len;
+    
+    cmk_str = base64_decode(cmk_base64);
+    cmk_len = cmk_str.size();
+    ciphertext_str = base64_decode(ciphertext_base64);
+    ciphertext_len = ciphertext_str.size();
+    aad_len = strlen(aad);
```
yang8621, line 275: Have you tested strlen(NULL)?

yang8621: merge @265~275 for simplicity, e.g.
    string cmk_str = base64_decode(cmk_base64);
    int cmk_len = cmk_str.size();
    string ciphertext_str = base64_decode(ciphertext_base64);
    int ciphertext_len = ciphertext_str.size();
    int aad_len = strlen(aad);

Liang1XZhao: It is handled in nodejs. If the request parameter does not have aad, an empty string will be passed to napi

ehsm_kms_service/test/test_kms_with_rest.py
```diff
@@ -280,7 +282,7 @@ def get_args():
     port = args.port
     return ip, port
     
-def check_result(res, action, test_name):
+def check_respult(res, action, test_name):
```
yang8621, line 285: typo?

ehsm_kms_service/test/test_kms_with_rest.py
```diff
@@ -15,17 +14,20 @@
...
-
-    params["sign"] = sign
+    params["nonce"] = "".join(random.choice("0123456789") for i in range(16))
+    params["payload"] = urllib.parse.unquote(urllib.parse.urlencode(payload))
+    msec = str(int(time.time()*1000*1000%1000))
+    if(len(msec) == 1):
+        msec = '00' + msec
+    elif(len(msec) == 2):
+        msec = '0' + msec
+    params["timestamp"] = str(time.strftime("%Y%m%d%H%M%S", time.localtime())) + msec
```
yang8621, line 25: is timestamp really required? In js layer, the timestamp is generated by the server.

core/App/ehsm_napi.cpp
yang8621: NULL add is valid.

core/App/ehsm_provider.h
```diff
@@ -40,6 +40,10 @@
 
 #define EH_ENCRYPT_MAX_SIZE (6*1024)
 
+#define EH_AAD_MAX_SIZE (9*1024)
+
```
syan10, line 44: why not use 8K?

docker/build_and_run.sh
```diff
@@ -56,7 +59,7 @@ function run {
 	RUN_ARG="$RUN_ARG --device=/dev/sgx/enclave -v aesmd-socket:/var/run/aesmd"
 
 	# run the container
-	docker run -d $RUN_ARG -it -p 9000:9000 $EHSM_DOCKER_IMAGE_NAME
+	docker run -d $RUN_ARG -it -p $DOCKER_PORT:$HOST_PORT $EHSM_DOCKER_IMAGE_NAME
```
syan10, line 62: seems the map is reversed

ehsm_kms_service/ehsm_kms_params.js
```diff
@@ -14,134 +14,131 @@ const ehsm_keyorigin_t = {
...
+const cmk_base64 = {
+  type: 'string',
+  minLength: 1,
+  maxLength: MAX_LENGTH,
+  required: true,
+}
 const aad = {
   type: 'string',
   maxLength: 32,
+  required: false
```
syan10, line 28: doe the require can be "optional"?

ehsm_kms_service/user_info.js
```diff
@@ -0,0 +1,4 @@
+const user_info = {
+    "202112101919":"202112345678"
```
syan10, line 2: what's this for?

core/App/ehsm_napi.cpp
```diff
@@ -81,11 +81,11 @@ void NAPI_Finalize(){
 */
 char* NAPI_CreateKey(const uint32_t keyspec, const uint32_t origin)
 {
+    RetJsonObj retJsonObj;
     ehsm_status_t ret = EH_OK;
-    ehsm_keyblob_t master_key;
+    ehsm_keyblob_t master_key = {0};
```
syan10, line 87: you can add an open here, not sure if it worked for the nested struture initialization.

it's more better to call the memset.

core/App/ehsm_napi.cpp
```diff
@@ -462,26 +573,69 @@ char* NAPI_Sign(const char* cmk_base64,
...
+            break;
+        case EH_EC_SM2:
+            // TODO
+            break;
+        default:
+            retJsonObj.setCode(retJsonObj.CODE_BAD_REQUEST);
+            retJsonObj.setMessage("The cmk's keyspec is invalid.");
+            goto out;
+    }
+
```
syan10, line 627: don't think you need the switch() from line#605-line#627, because the sign operation only accept the parameter as digest which is generated for the original message by using a hash algorithm. The hash algorithm is specified by the Algorithm parameter.

in most cases the digest is 256 bytes. or not longer than it.

core/App/ehsm_napi.cpp
```diff
@@ -530,30 +684,83 @@ char* NAPI_Verify(const char* cmk_base64,
...
```
syan10, line 744: same issue with the sign()

core/App/ehsm_napi.cpp
```diff
@@ -582,26 +789,69 @@ char* NAPI_AsymmetricEncrypt(const char* cmk_base64,
...
```
syan10, line 843: for the AsymmetricDecrypt/AsymmetriEncrypt, the above check is actually need.

core/App/ehsm_napi.cpp
Liang1XZhao: It will be fix.

ehsm_kms_service/test/test_kms_with_rest.py
Liang1XZhao: It will be fix.

core/App/ehsm_provider.h
Liang1XZhao: It will be fix.

core/App/ehsm_napi.cpp
Liang1XZhao: It will be change.

docker/build_and_run.sh
Liang1XZhao: It will be fix.

ehsm_kms_service/test/test_kms_with_rest.py
Liang1XZhao: It will be fix.

core/App/ehsm_napi.cpp
Liang1XZhao: It will be fix.

core/App/ehsm_napi.cpp
Liang1XZhao: It will be fix

ehsm_kms_service/ehsm_kms_params.js
Liang1XZhao: yes

ehsm_kms_service/user_info.js
Liang1XZhao: This is a temporary database used to obtain appkey matching appid. Later, this content will be migrated to the database

ehsm_kms_service/test/test_kms_with_rest.py
```diff
@@ -15,17 +14,20 @@
...
+    sign_string = urllib.parse.unquote(urllib.parse.urlencode(params))
+    sign_string += '&appkey=' + appkey
```
yang8621, line 27: for easy use, remove this line: sign_string += '&appkey=' + appkey
appkey has been used in HMAC-SHA256.

yang8621:
The replay attack can be successfully processed by setting 14mins faster than the server time:
1. Server time is 12:00, at the time, the server receives the RequestA.
2. The RequestA timestamp is 12:14(within 15min diff range), nonce is 1234. nonce is saved to nonce database with server time 12:00.
3. Server time @12:01, nonce database is checked and nonce1234 is kept.
4. Server time @12:16, nonce database is checked and nonce1234 is removed.
5. @12:17, an identical RequestA is replayed. The request can be processed because the server time 12:17 and the requestA timestamp 12:14 is in a tolerable difference.

ehsm_kms_service/ehsm_kms_server.js
```diff
@@ -4,10 +4,16 @@ const app = express();
...
 const ehsm_napi = require('./ehsm_napi');
 const { ehsm_kms_params, ehsm_keyspec_t, ehsm_keyorigin_t} = require('./ehsm_kms_params.js')
-
+const userInfo = require('./user_info')
 app.use(express.json());
 
-const port = 9000;
+const PORT = process.argv.slice(2)[0] || 9000;
+
+const NONCE_CACHE_TIME = 15 * 60 * 1000;
```
yang8621, line 12: As the replay attack I mentioned, the nonce should be kept in the database for 2*MAX_TIME_STAMP_DIFF.

ehsm_kms_service/ehsm_kms_server.js
```diff
@@ -37,69 +43,189 @@ const base64_encode = (str) => new Buffer.from(str).toString('base64');
...
+      }
+      if(slice_index == 0) {
+        delete nonce_database[n];
+      }
+    }
+  } catch (error) {
+    res.send(result(404, 'Not Fount', {}));
+    logger.info(JSON.stringify(error));
+  }
+}, NONCE_CACHE_TIME);
```
yang8621, line 136: here set the interval to: NONCE_CACHE_TIME/2

ehsm_kms_service/ehsm_kms_server.js
Liang1XZhao: the MAX_TIME_STAMP_DIFF will be changed to 5 minutes,It is a typo.

yang8621: explicitly show the relationship between MAX_TIME_STAMP_DIFF and NONCE_CACHE_TIME:
NONCE_CACHE_TIME = 2 * MAX_TIME_STAMP_DIFF

core/App/ehsm_napi.cpp
```diff
@@ -224,41 +270,78 @@ char* NAPI_Decrypt(const char* cmk_base64,
...
     ehsm_data_t plaint_data;
     ehsm_data_t aad_data;
     ehsm_data_t cipher_data;
+    string plaintext_base64;
+    
+    string cmk_str;
+    string ciphertext_str;
+    int cmk_len;
+    int ciphertext_len;
+    int aad_len;
```
syan10, line 294: seems you didn't update the codes here following yang's comments. and same with below codes.

ehsm_kms_service/ehsm_kms_params.js
syan10: so set it as optional, don't use false.

ehsm_kms_service/ehsm_kms_server.js
syan10: seems the logic is too complicated!

we can use the timestamp only. it can also prevent the replay attack.

ehsm_kms_service/test/test_kms_with_rest.py
Liang1XZhao: It will be fix

ehsm_kms_service/test/test_kms_with_rest.py
```diff
@@ -15,17 +14,14 @@
...
     params["appid"] = appid
-    params["nonce"] = str(int.from_bytes(os.urandom(16), "big"))
-    params["timestamp"] = int(time.time())
-
-    query_string = urllib.parse.urlencode(params)
-    query_string += '&app_key=' + appkey
-    sign = str(base64.b64encode(hmac.new(appkey.encode('utf-8'), query_string.encode('utf-8'), digestmod=sha256).digest()),'utf-8')
-
-    params["sign"] = sign
+    params["nonce"] = "".join(random.choice("0123456789") for i in range(16))
```
yang8621, line 18: remove nonce, use timestamp instead:
params["timestamp"] = str(int(time.time() * 1000000))

ehsm_kms_service/ehsm_kms_server.js
```diff
@@ -38,68 +43,199 @@ const base64_encode = (str) => new Buffer.from(str).toString('base64');
...
+    if(!_checkPayload_res){
+      return;
+    };
+    
+    // check sign
+    let appkey = userInfo[appid];
+    let sign_parmas = { appid, timestamp, payload }
+    let sign_string = params_sort_str(sign_parmas);
+    let local_sign = crypto.createHmac('sha256', appkey).update(sign_string, 'utf8').digest('base64');
+    if(sign != local_sign) {
```
yang8621, line 222: what's the result:   (AbC == aBc),  true or false ?

yang8621 (approved):

syan10 (approved):

PR #71, opened 2022-02-14
fix the nonce cache issue

panpan0721:
The latest nonce cache data will be always added at the beginning
of the database array. like :[new nonceData, old nonceData, ...].

The slice_index points to the index of the first expired nonce cache data.
Previously, it use the slice(slice_index) to return the updated array which only
contains the cache data that already expired.

So change to use slice(0, slice_index) to return the updated array that contains the
non-expired cache data.

### DIFF
--- a/ehsm_kms_service/function.js
+++ b/ehsm_kms_service/function.js
@@ -37,13 +37,17 @@ const base64_decode = (base64_str) =>
 
 /**
  * Clear nonce cache for more than <NONCE_CACHE_TIME> minutes
- * @returns nonce_cache_timer
+ * nonce_database[appid]
+ *  - type: array
+ *  - sort: [new timestamp, old timestamp, ...]
+ * @returns nonce_cache_timer, nonce_database
  */
 const _nonce_cache_timer = () => {
   const nonce_database = {}
   const timer = setInterval(() => {
     try {
       for (const appid in nonce_database) {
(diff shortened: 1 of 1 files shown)


ehsm_kms_service/function.js
```diff
@@ -37,13 +37,17 @@ const base64_decode = (base64_str) =>
...
  */
 const _nonce_cache_timer = () => {
   const nonce_database = {}
   const timer = setInterval(() => {
     try {
       for (const appid in nonce_database) {
+        // slice_index  Index of the cache that exceeded the maximum time
         let slice_index =
           nonce_database[appid] &&
           nonce_database[appid].findIndex((nonce_data) => {
```
yang8621, line 53: Will findIndex() return one or all matched results?

panpan0721: Returns the position of the first element of the qualified array

yang8621 (approved):

ehsm_kms_service/function.js
```diff
@@ -37,13 +37,17 @@ const base64_decode = (base64_str) =>
 
 /**
```
syan10, line 38: can you add more explanation for the issue and the fix solution in the commit msg?

panpan0721: ok

yang8621 (approved):

syan10 (approved):

PR #80, opened 2022-02-24
pass the dkeyserver addr as input params

wanghouqi:
remove hand code ip and port, change to passed in by the startup
parameter [-h] and [-p]

Signed-off-by: wanghouqi <wanghouqi@vip.qq.com>

### DIFF
--- a/dkeycache/App/main.cpp
+++ b/dkeycache/App/main.cpp
@@ -3,6 +3,8 @@
 #include <signal.h>
 
 #include <enclave_u.h>
+#include <getopt.h>
+#include "log_utils.h"
 
 // Need to create enclave and do ecall.
 #include "sgx_urts.h"
@@ -31,6 +33,13 @@ void ocall_print_string(const char *str)
 LaTask * g_la_task = NULL;
 LaServer * g_la_server = NULL;
 
+std::string deploy_ip_addr;
+uint32_t deploy_port = 0;
+static const char* _sopts = "i:p:";
+static const struct option _lopts[] = {{"ip", required_argument, NULL, 'i'},
+                                       {"port", required_argument, NULL, 'p'},
+                                       {0, 0, 0, 0}};
+
 void signal_handler(int sig)
 {
     switch(sig)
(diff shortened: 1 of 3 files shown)


dkeycache/App/main.cpp
```diff
@@ -59,8 +59,27 @@ void cleanup()
...
+        if (argv[i] == host_key && i + 1 < argc && argv[i + 1][0] != '-') {
+            host_index = i + 1;
+        } else if (argv[i] == port_key && i + 1 < argc && argv[i + 1][0] != '-') {
+            port_index = i + 1;
+        }
+    }
+    if (host_index == -1 || port_index == -1)
+    {
+        printf("\nusage: ehsm-dkeycache -h 127.0.0.1 -p 8888\n\n");
+        return -1;
```
syan10, line 79: you can follow the enroll_app, don't use "-h" or "-p"
just directly run as "ehsm-dkeycache <dkeyserver_url>" and the dkeyserver_url="127.0.0.1:8888"

syan10: ignore it, seems you need to parse the input to get the IP and port

wanghouqi: It is fixed.please check again.

dkeycache/App/main.cpp
```diff
@@ -57,10 +66,44 @@ void cleanup()
         delete g_la_server;
 }
 
+static void show_usage_and_exit(int code) {
+    printf("\nusage: ehsm-dkeycache -h 127.0.0.1 -p 8888\n\n");
+    exit(code);
```
syan10, line 71: "-h" -> "-i"

"-h" for help, to print the usage

dkeycache/App/main.cpp
```diff
@@ -3,6 +3,8 @@
 #include <signal.h>
 
 #include <enclave_u.h>
+#include <getopt.h>
```
syan10, line 6: change the commit title as 
“
pass the dkeyserver addr as input params
”

dkeycache/App/main.cpp
wanghouqi: ok,It will be change.

dkeycache/App/main.cpp
wanghouqi: changed

yang8621 (approved):

PR #86, opened 2022-03-07
add implementation for NAPI_GenerateQuote and NAPI_VerifyQuote

syan10:
add a new native unittest for the APIs.

Signed-off-by: Yan, Shaopu <shaopu.yan@intel.com>

### DIFF
--- a/core/App/ehsm_napi.cpp
+++ b/core/App/ehsm_napi.cpp
@@ -1408,7 +1408,7 @@ OUT:
 char* NAPI_GenerateQuote(const char *challenge)
 {
     RetJsonObj retJsonObj;
-    log_d("***NAPI_GenerateQuote start.");
+
     if (challenge == NULL) {
         retJsonObj.setCode(retJsonObj.CODE_BAD_REQUEST);
         retJsonObj.setMessage("paramter invalid.");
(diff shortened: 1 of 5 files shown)


core/App/ehsm_napi.cpp
```diff
@@ -1446,29 +1472,42 @@ char* NAPI_GenerateQuote(const char *challenge)
...
+    sgx_ql_qv_result_t verifyresult;
+    bool result = false;
+
+    ehsm_data_t quote;
+    memset(&quote, 0, sizeof(quote));
 
-    // TODO: Verify Quote
-    result = true;
+    string quote_str = base64_decode(quote_base64);
+    quote.datalen = quote_str.size();
```
yang8621, line 1493: quote_str may be NULL?

core/App/ehsm_napi.cpp
```diff
@@ -1446,29 +1472,42 @@ char* NAPI_GenerateQuote(const char *challenge)
  *          }
  *      }
  */
-char* NAPI_VerifyQuote(const char *quote, const char *nonce){
+char* NAPI_VerifyQuote(const char *quote_base64, const char *nonce)
+{
```
yang8621, line 1476: where check quote is base64 encoding? in node.js or add to this function?

syan10: yes, checked in js layer

core/App/ehsm_napi.cpp
syan10: fixed

yang8621 (approved):

PR #88, opened 2022-03-07
Modify base64 encoding feature about some input paraments

panpan0721:
1. Add base64 format checking for aad and plaintext paraments
2. Remove all api keys' "_base64" suffix

fix the corresponding test cases
test: passed the unittest

### DIFF
--- a/ehsm_kms_service/ehsm_kms_params.js
+++ b/ehsm_kms_service/ehsm_kms_params.js
@@ -19,11 +19,11 @@ const keyid = {
   type: 'string',
   minLength: 1,
   maxLength: MAX_LENGTH,
-  required: true,
+  required: false,
 }
 const aad = {
-  type: 'string',
-  maxLength: 32,
+  type: 'base64',
+  maxLength: MAX_LENGTH,
   required: false,
 }
 
@@ -43,9 +43,12 @@ const ehsm_kms_params = {
     },
   },
   Encrypt: {
-    keyid,
+    keyid: {
+      ...keyid, 
+      required: true
+    },
     plaintext: {
-      type: 'string',
+      type: 'base64',
       maxLength: MAX_LENGTH,
       minLength: 1,
       required: true,
@@ -53,9 +56,12 @@ const ehsm_kms_params = {
     aad,
   },
   Decrypt: {
-    keyid,
+    keyid: {
+      ...keyid, 
+      required: true
+    },
     ciphertext: {
-      type: 'string',
+      type: 'base64',
       maxLength: MAX_LENGTH,
       minLength: 1,
       required: true,
@@ -63,7 +69,10 @@ const ehsm_kms_params = {
     aad,
   },
   GenerateDataKey: {
-    keyid,
+    keyid: {
+      ...keyid, 
+      required: true
+    },
     keylen: {
       type: 'int',
       maxNum: 1024,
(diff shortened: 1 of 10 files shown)


syan10 (approved):

yang8621 (changes requested):
wait for your more changes

ehsm_kms_service/function.js
```diff
@@ -34,6 +34,7 @@ const base64_encode = (str) => new Buffer.from(str).toString('base64')
 // base64 decode
 const base64_decode = (base64_str) =>
   new Buffer.from(base64_str, 'base64').toString()
+const is_base64 = (base64_str) => /^[a-zA-Z0-9\+\=\/]+$/gi.test(base64_str);
```
yang8621, line 37: length is 4X, (X > 0)  
= and == are at the ending

you could search from google.

ehsm_kms_service/ehsm_kms_params.js
```diff
@@ -22,7 +22,7 @@ const keyid = {
   required: true,
 }
```
yang8621, line 23: I do hope keyid has its own reg check.
we could support it in others patches.

ehsm_kms_service/function.js
```diff
@@ -34,6 +34,15 @@ const base64_encode = (str) => new Buffer.from(str).toString('base64')
 // base64 decode
 const base64_decode = (base64_str) =>
   new Buffer.from(base64_str, 'base64').toString()
+const is_base64 = (base64_str) => {
+  if (base64_str == undefined ||
+      typeof base64_str != 'string' ||
+      base64_str.length == 0 ||
```
yang8621, line 40: const aad = {
  type: 'base64',
  maxLength: MAX_LENGTH,
  required: false,
}

aad='' is valid

ehsm_kms_service/function.js
```diff
@@ -333,6 +342,26 @@ const _checkPayload = function (req, res) {
           return false
         }
       }
+      if (currentPayLoad[key].type == 'base64' && payload[key]) {
```
yang8621, line 345: why not "else if" in _checkPayload

ehsm_kms_service/function.js
```diff
@@ -333,6 +342,26 @@ const _checkPayload = function (req, res) {
...
+        if (
+          payload[key] != undefined &&
+          ((currentPayLoad[key].maxLength &&
+            payload[key].length > currentPayLoad[key].maxLength) ||
+            (currentPayLoad[key].minLength &&
+              payload[key].length < currentPayLoad[key].minLength))
+        ) {
+          res.send(_result(400, `${key} length error`))
+          return false
+        }
```
yang8621, line 359: Can we special handle aad='' here like:

if (!currentPayLoad[key].minLength && payload[key].length === 0)
return true

not sure, please confirm.

yang8621:
one curiosity:
https://github.com/intel/ehsm/blob/2ffb7e4301ae55ba5884ef3a689cfcf76c622d4d/ehsm_kms_service/function.js#L321
what will happen if currentPayLoad is undefined.

ehsm_kms_service/function.js
panpan0721: This is a common function for checking valid base64, not only for aad. 
And "required: false," will cover the case that "aad=''" at line 322 - 328

ehsm_kms_service/function.js
panpan0721: I just try to follow the code  style like "string", "int" and else other

I dont known why the checking code dont use "else if" before too - -///.

ehsm_kms_service/function.js
panpan0721: This case has been covered at line 322 - 328

yang8621 (approved):

PR #92, opened 2022-03-08
Change pccs server url into environmental variables

panpan0721:
Remove sgx_default_qcnl.conf file generating in Dockfile

Add sgx_config.sh shell script for getting pccs url configration
in environmental and generating sgx_default_qcnl.conf on
docker start.

Add start.sh shell script for calling sgx_config.sh and the docker
work payload command.


docker/core/start.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin sh
+sh /home/sgx_config.sh
+node /home/ehsm/ehsm_kms_service/ehsm_kms_server.js
```
syan10, line 3: combine the twi scripts into one, (named as start_ehsm-kms.sh)

docker/dkeycache/start.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin sh
+sh /home/sgx_config.sh
```
syan10, line 2: combine and named as "start_dkeycache.sh"

docker/dkeyserver/start.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin sh
+sh /home/sgx_config.sh
+sh -c /home/ehsm/out/ehsm-dkeyserver/ehsm-dkeyserver
```
syan10, line 3: combine and named as start_dkeyserver.sh

panpan0721: It will be fixed

docker/dkeycache/start.sh
panpan0721: It will be fixed

docker/core/start.sh
panpan0721: It will be fixed

syan10 (approved):

PR #95, opened 2022-03-10
Fix test_kms_with_rest errors\

HawieLong:
1. fix the upper used in test_kms_with_rest which cause base64 error
2. fix disable key coding error

Signed-off-by: Long Hanyu <hanyu.long@intel.com>

### DIFF
--- a/ehsm_kms_service/test/test_kms_with_rest.py
+++ b/ehsm_kms_service/test/test_kms_with_rest.py
@@ -71,7 +71,7 @@ def test_params(payload):
     sign_string = urllib.parse.unquote(urllib.parse.urlencode(params))
     print(sign_string.encode('utf-8'))
     print(apikey.encode('utf-8'))
-    sign = str(base64.b64encode(hmac.new(apikey.encode('utf-8'), sign_string.encode('utf-8'), digestmod=sha256).digest()),'utf-8').upper()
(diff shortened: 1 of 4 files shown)


ehsm_kms_service/test/test_kms_with_rest.py
```diff
@@ -177,7 +177,7 @@ def test_Stest_RSA3072_sign_verify(base_url, headers):
 
     # test Sign
     payload.clear()
-    payload["digest"] = str(base64.b64encode("test".encode("utf-8")),'utf-8').upper()
+    payload["digest"] = str(base64.b64encode("test".encode("utf-8")),'utf-8')
```
yang8621, line 180: nice fix. 
but still same issue @line 217 229  255  267

yang8621:
Please also fix: https://github.com/intel/ehsm/blob/main/ehsm_kms_service/test/cli/_utils_.py#L28

ehsm_kms_service/test/test_kms_with_rest.py
HawieLong: the sign will fix later, since the sign is aligned with server, the sever is using upper now.

yang8621: Seems: https://github.com/intel/ehsm/blob/main/ehsm_kms_service/function.js#L603

yang8621 (approved):

syan10 (approved):

PR #111, opened 2022-03-25
Add docker compose support

wanghouqi:
test: passed the unittest

Signed-off-by: wanghouqi <wanghouqi@vip.qq.com>


docker/docker-compose.yml
```diff
@@ -0,0 +1,102 @@
+version: "3.9"
+services:
```
syan10, line 2: this version is for compose version? please add a comment

docker/.env
```diff
@@ -0,0 +1,29 @@
...
+COUCHDB_DB=ehsm_kms_db
+
+# docker image name
+IMAGE_NAME_EHSM_BASE=ehsm_base
+
+# eHSM version
+vsn_major=0
+vsn_minor=2
+vsn_patch=0
+EHSM_VERSION_STR=${vsn_major}.${vsn_minor}.${vsn_patch}
```
syan10, line 17: is there some way to get this version automatically?

docker/.env
```diff
@@ -0,0 +1,29 @@
+# host ip
```
syan10, line 1: update the build-instructions doc to install the dependencies for the docker-compose

docker/docker-compose.yml
wanghouqi: no,It is docker compose file version.
https://docs.docker.com/compose/compose-file/

docker/.env
wanghouqi: ok

docker/.env
wanghouqi: About the version,We need to discuss it .

syan10 (approved):

yang8621 (approved):

HawieLong (approved):

PR #124, opened 2022-04-18
Integrate SGX SSL as a third party library

yang8621:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2011-2020 Intel Corporation. All rights reserved.
+# Copyright (C) 2011-2022 Intel Corporation. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -54,7 +63,31 @@ else ifeq ($(Build_Mode), SIM_PRERELEAESE)
 	@echo "The project has been built in simulation pre-release mode."
 endif
 
+ssl:
```
syan10, line 66: can we move the this into the sub folder of SSL?

buildenv.mk
```diff
@@ -69,6 +69,16 @@ CC ?= gcc
 CXX ?= g++
 RM = rm -f
 
+######## SGX SSL Settings ########
```
syan10, line 72: same here, it's specific for ssl

core/App/ehsm_provider.cpp
```diff
@@ -958,7 +958,7 @@ ehsm_status_t VerifyQuote(ehsm_data_t *quote,
     // e.g. You can get latest QvE ISVSVN in QvE Identity JSON file from
     // https://api.trustedservices.intel.com/sgx/certification/v2/qve/identity
     // Make sure you are using trusted & latest QvE ISV SVN as threshold
-    sgx_isv_svn_t qve_isvsvn_threshold = 3;
+    sgx_isv_svn_t qve_isvsvn_threshold = 5;
```
syan10, line 962: why change it?

yang8621: verify quote failed with the error:
ERROR [App/ehsm_provider.cpp(1048) -> VerifyQuote]: Error in Verify QvE report and identity failed. 0xe051

syan10 (approved):

PR #130, opened 2022-05-05
Https support

yang8621:
Add https support
A self-signed cetificate is taken as the example
but only used for non-production environment.
Http request is not supported anymore.


yang8621:
hold review. not fully done yet.

syan10 (commented):
LGTM, but suggest to add more details description in the commit msg.

enroll_app/rest_utils.cpp
```diff
@@ -79,8 +87,8 @@ std::string post(std::string url, std::string content)
 /*
  * KMS server post
  */
-void post_KMS(std::string url, std::string content, RetJsonObj *retJsonObj)
+void post_KMS(std::string url, std::string content, RetJsonObj *retJsonObj, bool secure)
```
syan10, line 90: "secure" may lead some confuese, suggest to change it as verify? or something better words

syan10 (approved):

PR #135, opened 2022-07-20
add init defaultCMK for KMS service

Yidingluo1:
When the KMS service is started, the defaultCMK will be created, if it
has been created before, success will be returned.
test: passed the test

Signed-off-by: luoyiding <luo.yd@neusoft.com>


ehsm_kms_service/secret_management.js
```diff
@@ -0,0 +1,64 @@
...
+                    const napi_res = JSON.parse(ehsm_napi[`NAPI_${cryptographic_apis.CreateKey}`](keyspec, origin))
+                    const creationDate = new Date().getTime()
+                    DB.insert({
+                        _id: `cmk:${keyid}`,
+                        keyid,
+                        keyBlob: napi_res.result.cmk,
+                        creator: 'System',
+                        creationDate,
+                        expireTime: creationDate + CMK_EFFECTIVE_DURATION * 999,
+                        alias: 'System Default CMK',
```
syan10, line 41: 'Secret Manager Default CMK'

ehsm_kms_service/secret_management.js
```diff
@@ -0,0 +1,64 @@
...
+                if (cmks_res.docs.length == 0) {
+                    const keyspec = 0
+                    const origin = 0
+                    const napi_res = JSON.parse(ehsm_napi[`NAPI_${cryptographic_apis.CreateKey}`](keyspec, origin))
+                    const creationDate = new Date().getTime()
+                    DB.insert({
+                        _id: `cmk:${keyid}`,
+                        keyid,
+                        keyBlob: napi_res.result.cmk,
+                        creator: 'System',
```
syan10, line 38: you can change the "System" to "Secret Manager"

ehsm_kms_service/constant.js
```diff
@@ -6,6 +6,7 @@ const CMK_EFFECTIVE_DURATION = 1 * 365 * 24 * 60 * 60 * 1000
 const CMK_LOOP_CLEAR_TIME = 24 * 60 * 60 * 1000
 const CMK_EXPIRE_TIME_EXPAND = 10 * 24 * 60 * 60 * 1000
 const CMK_LOOP_CLEAR_EXECUTION_TIME = 3
+const KEYID_SYSTEM_DEFAULT_CMK = 'systemDefCMK'
```
syan10, line 9: same here, SYSTEM to SecretManager

yang8621, line 9: Def means define, change to sysDftCMK or others

ehsm_kms_service/ehsm_kms_server.js
```diff
@@ -15,6 +15,9 @@ const {
   _nonce_cache_timer,
   _cmk_cache_timer,
 } = require('./function')
+const {
+  init_DefaultCMK,
```
syan10, line 19: Init_SM_DefaultCMK

ehsm_kms_service/secret_management.js
```diff
@@ -0,0 +1,64 @@
...
+    try {
+        const keyid = KEYID_SYSTEM_DEFAULT_CMK
+        const query = {
+            selector: {
+                keyid,
+            },
+            fields: ['keyid',],
+            limit: 1,
+        }
+        DB.partitionedFind('cmk', query)
```
yang8621, line 27: can reuse store_cmk() ?

Yidingluo1: Parameters and return values are different, if reused，store_cmk() needs to be modified

yang8621: OK.

yang8621:
We need separate cmk for different users, but not a default key for all users which introduce security concern.

Yidingluo1:
> We need separate cmk for different users, but not a default key for all users which introduce security concern.

When creating secret, the default CMK will be used only if the user does not fill in the CMK. This default CMK is only called when the service is started, and it is only a default value when there is no CMK to type

yang8621:
> > We need separate cmk for different users, but not a default key for all users which introduce security concern.
> 
> When creating secret, the default CMK will be used only if the user does not fill in the CMK. This default CMK is only called when the service is started, and it is only a default value when there is no CMK to type

So is it possible that two users share the same default CMK? The default CMK still requires being different for different users. That means default CMK should bind to appid.

Yidingluo1:
Scheme abandonment

PR #137, opened 2022-07-21
add init default SM CMK

Yidingluo1:
add default CMK when enroll user info

Signed-off-by: luoyiding <luo.yd@neusoft.com>

### DIFF
--- a/ehsm_kms_service/function.js
+++ b/ehsm_kms_service/function.js
@@ -264,8 +264,11 @@ const enroll_user_info = (action, DB, res, req) => {
   if (napi_res) {
     const { appid, apikey } = napi_res.result
     let cmk_res = napi_result(cryptographic_apis.CreateKey, res, [0, 0])
-    if (cmk_res) {
+    let sm_default_cmk_res = napi_result(cryptographic_apis.CreateKey, res, [0, 0])
+    if (cmk_res && sm_default_cmk_res) {
       const { cmk } = cmk_res.result
+      // create a default secret manager CMK for current appids
+      const sm_default_cmk = sm_default_cmk_res.result.cmk
       let apikey_encrypt_res = napi_result(cryptographic_apis.Encrypt, res, [
         cmk,
         apikey,
@@ -278,6 +281,7 @@ const enroll_user_info = (action, DB, res, req) => {
           appid,
           apikey: ciphertext,
           cmk: cmk,
+          sm_default_cmk,
         })
           .then((r) => {
             res.send(_result(200, 'successful', { ...napi_res.result }))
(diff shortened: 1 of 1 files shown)


ehsm_kms_service/function.js
```diff
@@ -264,8 +264,10 @@ const enroll_user_info = (action, DB, res, req) => {
   if (napi_res) {
     const { appid, apikey } = napi_res.result
     let cmk_res = napi_result(cryptographic_apis.CreateKey, res, [0, 0])
+    let sm_default_cmk_res = napi_result(cryptographic_apis.CreateKey, res, [0, 0])
```
yang8621, line 267: will the result be checked?

Yidingluo1: Yes, it has been tested

ehsm_kms_service/function.js
```diff
@@ -264,8 +264,10 @@ const enroll_user_info = (action, DB, res, req) => {
   if (napi_res) {
     const { appid, apikey } = napi_res.result
     let cmk_res = napi_result(cryptographic_apis.CreateKey, res, [0, 0])
+    let sm_default_cmk_res = napi_result(cryptographic_apis.CreateKey, res, [0, 0])
     if (cmk_res) {
```
yang8621, line 268: Does the below change make sense?
if (sm_default_cmk_res && cmk_res)

Yidingluo1: > Does the below change make sense? if (sm_default_cmk_res && cmk_res)

I think if you want to make cmk different from defaultcmk, you must write it separately

wanghouqi: the if (cmk_res) will be changed to if (sm_default_cmk_res && cmk_res)

yang8621 (approved):

ehsm_kms_service/function.js
syan10: generally, the abbreviation for "result" is "ret" not "res"!

anyway, you can change it later.

ehsm_kms_service/function.js
```diff
@@ -264,8 +264,10 @@ const enroll_user_info = (action, DB, res, req) => {
   if (napi_res) {
     const { appid, apikey } = napi_res.result
     let cmk_res = napi_result(cryptographic_apis.CreateKey, res, [0, 0])
-    if (cmk_res) {
+    let sm_default_cmk_res = napi_result(cryptographic_apis.CreateKey, res, [0, 0])
+    if (cmk_res && sm_default_cmk_res) {
       const { cmk } = cmk_res.result
+      const sm_default_cmk = sm_default_cmk_res.result.cmk
```
syan10, line 270: add some comments here

wanghouqi: will be add: // create a default secret manager CMK for current appid.

syan10 (approved):

PR #147, opened 2022-08-03
add support for the putSecretValue restful API

Yidingluo1:
this restful API will be exposed to public as:
POST <ehsm_srv_address>/ehsm?Action=PutSecretValue

stores the secret value of a new version into a secret object.
by default, the newly stored secret value is marked with the current version stage, and the original versions
will be set as previously.

test: passed the unittest


ehsm_kms_service/secret_manager_apis.js
```diff
@@ -265,9 +265,104 @@ const createSecret = async (res, appid, payload, DB) => {
...
+                'createTime',
+                'versionStage',
+            ],
+            limit: 1
+        }
+        const version_stage_res = await DB.partitionedFind('secret_version_data', query_version_stage)
+        if (version_stage_res.docs.length > 0) {
+            versionId = version_stage_res.docs[0].versionId + 1
+            version_stage_res.docs[0].versionStage = SM_SECRET_VERSION_STAGE_PREVIOUS
+            await DB.insert(version_stage_res.docs[0])
```
HawieLong, line 338: use "insert" to update a DB? Can we use “update DB“ code here？

ehsm_kms_service/secret_manager_apis.js
```diff
@@ -265,9 +265,104 @@ const createSecret = async (res, appid, payload, DB) => {
...
+        if (secret_metadata_res.docs.length == 0) {
+            encryptionKeyId = ""
+        } else {
+            encryptionKeyId = secret_metadata_res.docs[0].encryptionKeyId
+        }
+        retCMK = await get_SM_Masterkey(DB, appid, encryptionKeyId)
+        if (retCMK.isFind) {
+            sm_masterKey = retCMK.sm_masterKey
+        } else {
+            res.send(_result(400, retCMK.msg))
```
HawieLong, line 306: If the user uses a new secretName to put, we still return CMK not find? We should confirm the secretName in DB first.

ehsm_kms_service/secret_manager_apis.js
```diff
@@ -265,9 +265,104 @@ const createSecret = async (res, appid, payload, DB) => {
...
+        const query_encryptionKeyId = {
+            selector: {
+                appid,
+                secretName,
+            },
+            fields: ['encryptionKeyId'],
+            limit: 1,
+        }
+        const secret_metadata_res = await DB.partitionedFind('secret_metadata', query_encryptionKeyId)
+        if (secret_metadata_res.docs.length == 0) {
```
yang8621, line 297: if secretName does not exist (CreateSecret has not ever been called before), the code will go into @298?

ehsm_kms_service/secret_manager_apis.js
Yidingluo1: no direct update function in CouchDB, can only be replaced by insert

ehsm_kms_service/secret_manager_apis.js
Yidingluo1: no direct update function in CouchDB, can only be replaced by insert

ehsm_kms_service/secret_manager_apis.js
Yidingluo1: problem fixed. now query the secretname first, and then find the encryptionKeyId exists

ehsm_kms_service/secret_manager_apis.js
Yidingluo1: problem fixed. now query the secretname first, and then find the encryptionKeyId exists

syan10:
new commti msg
"
add support for the putSecretValue restful API

this restful API will be exposed to public as:
POST <ehsm_srv_address>/ehsm?Action=PutSecretValue

stores the secret value of a new version into a secret object.
by default, the newly stored secret value is marked with the current version stage, and the original versions
will be set as previously.

test: passed the unittest
"

Yidingluo1:
> new commti msg " add support for the putSecretValue restful API
> 
> this restful API will be exposed to public as: POST <ehsm_srv_address>/ehsm?Action=PutSecretValue
> 
> stores the secret value of a new version into a secret object. by default, the newly stored secret value is marked with the current version stage, and the original versions will be set as previously.
> 
> test: passed the unittest "

commit changed

test/test_secret_manager_with_cli.py
```diff
@@ -30,6 +30,8 @@ def test_secret_manager(base_url, headers):
 
     createSecret.createSecret(base_url, "secretData2", "secret002", None, "mysecret", "20d")
 
+    putSecretValue.putSecrectValue(base_url, "secret001", "secretData3")
```
syan10, line 33: seems the param order not make sense, generally, if I'm a user for the secret, i will use it by the following cmd:

putSecretVaulle --secretName "xxxx" --secretData "xxx"

same with other APIs

Yidingluo1: ok, I will submit a new PR later to modify the parameter order of the create secret test

ehsm_kms_service/secret_manager_apis.js
```diff
@@ -265,9 +265,119 @@ const createSecret = async (res, appid, payload, DB) => {
...
+        }
+        retCMK = await get_SM_Masterkey(DB, appid, encryptionKeyId)
+        if (retCMK.isFind) {
+            sm_masterKey = retCMK.sm_masterKey
+        } else {
+            res.send(_result(400, retCMK.msg))
+            return
+        }
+        const apikey_encrypt_res = napi_result(cryptographic_apis.Encrypt, res, [sm_masterKey, secretData, ''])
+        const { ciphertext } = apikey_encrypt_res.result
```
yang8621, line 325: will check any error returned from Encrypt?

ehsm_kms_service/secret_manager_apis.js
yang8621: if _id is matched, DB.insert will update the same record?  is my understanding correct?

ehsm_kms_service/secret_manager_apis.js
```diff
@@ -265,9 +265,119 @@ const createSecret = async (res, appid, payload, DB) => {
...
+                '_rev',
+                'appid',
+                'secretName',
+                'versionId',
+                'deletedFlag',
+                'secretData',
+                'createTime',
+                'versionStage',
+            ],
+            limit: 1
```
yang8621, line 347: How can make sure the only one result is the latest versionid?

ehsm_kms_service/secret_manager_apis.js
yang8621: Line 353 is to insert a new record OR to update the original one?

ehsm_kms_service/secret_manager_apis.js
Yidingluo1: always only one secret_version_data version stage is SM_SECRET_VERSION_STAGE_PREVIOUS

ehsm_kms_service/secret_manager_apis.js
Yidingluo1: will catch in line 370

PR #150, opened 2022-08-04
add support for the deleteSecret restful API

Yidingluo1:
this restful API will be exposed to public as:
POST <ehsm_srv_address>/ehsm?Action=DeleteSecret

If you call this operation without specifying a recovery period, the deleted secret can be recovered within 30 days.
If you specify a recovery period, the deleted secret can be recovered within the recovery period. You can also forcibly delete a secret. A forcibly deleted secret cannot be recovered.

test: passed the unittest


ehsm_kms_service/secret_manager_apis.js
```diff
@@ -265,9 +314,124 @@ const createSecret = async (res, appid, payload, DB) => {
     }
 }
 
+//Force delete secret or schedule a time to delete secret
```
yang8621, line 317: I'd like to suggest adding parameter descriptions of payload for each secret manager's APIs.
secretName       Required    Length(1~100)    
recoveryPeriod  "brief introduction of the item if it's not straightforward"
Optional     Range(0~365)  Uint(ms/s/min/hour/day/month?)  DefaultValue(30d)   
forceDelete       "xxxxxxxx"
Optional    Range(T/F)        DefaultValue(False)

ehsm_kms_service/secret_manager_apis.js
```diff
@@ -151,6 +152,60 @@ function calculateRotationSperate(rotationInterval, createTime) {
...
+                        appid,
+                        secretName
+                    },
+                    fields: ['_id', '_rev'],
+                    limit: 1000
+                }
+                let secret_version_data_res = await DB.partitionedFind('secret_version_data', query_secret_version_data)
+                if (secret_version_data_res.docs.length > 0) {
+                    for (const secret_version_data_item of secret_version_data_res.docs) {
+                        secret_version_data_item._deleted = true
```
yang8621, line 190: forceDeleteData only delete metadata but not version data?

Yidingluo1: it will delete metadata and all version data by secretname and appid

ehsm_kms_service/secret_manager_apis.js
wanghouqi: will be add description for function

PR #157, opened 2022-08-11
add delete_secret_thread

Yidingluo1:
it will scan the database every 6 hours and delete items that have reached the deletion time.

passed the python test Signed-off-by: yidingluo1 [luo.yd@neusoft.com](mailto:luo.yd@neusoft.com)

### DIFF
--- a/ehsm_kms_service/delete_secret_thread.js
+++ b/ehsm_kms_service/delete_secret_thread.js
@@ -1,0 +1,118 @@
+const {
+    base64_decode
+} = require('./function')
+
+// Set a queue to record the waiting time
+var delete_task_queue = {}
+
+/**
+ * remove a timer from exist timer by key
+ * @param {Object} DB : database control
+ * @param {Object} key : key form exist timer
+ * @returns
+ */
+const remove_delete_task = (secretName, appid) => {
+    try {
+        //Delete thread when thread exists
+        if (delete_task_queue[secretName + '_' + appid] != undefined) {
+            clearTimeout(delete_task_queue[secretName + '_' + appid].timer)
+        }
+        delete delete_task_queue[secretName + '_' + appid];
+    } catch (error) {
+        console.error('error remove_delete_task :: ', error)
+    }
+}
+
+/**
+ * public interface, add a new delete task into the delete queue with the secret name.
+ * @param {String} appid : appid of user
+ * @param {Object} DB : database control
+ * @param {Object} secretName : The name of the secret. eg. 'secretName01'
+ * @param {Object} plannedDeleteTime : planned delete time for secret eg. '1659931644416'
+ * @returns
+ */
+const add_delete_task = (DB, appid, secretName, plannedDeleteTime) => {
+    try {
+        let delayMillisecond = plannedDeleteTime - new Date().getTime()
+        let needAdd = true
+        let key = secretName + '_' + appid
+        // process exist timer, 
+        // If the new time is within plus or minus 10 seconds of the original time, keep the original plan unchanged; otherwise, clear the original plan and add again
+        if (delete_task_queue.hasOwnProperty(key)) {
+            if (delete_task_queue[key].plannedDeleteTime + 10000 > plannedDeleteTime && delete_task_queue[key].plannedDeleteTime - 10000 < plannedDeleteTime) {
+                needAdd = false
+            } else {
+                clearTimeout(delete_task_queue[key].timer)
+            }
+        }
+        //If the planned delete time exceeds 12 hours, the thread will not be started
+        if (delayMillisecond > 12 * 1000 * 60 * 60) {
+            needAdd = false
+        }
+        if (needAdd) {
+            //Force delete according to planned deletion time
+            let timer = setTimeout(async () => {
+                try {
+                    const forceDeleteData = require('./secret_manager_apis').forceDeleteData
+                    if (await forceDeleteData(DB, appid, secretName)) {
+                        console.info(`delete secret thread :: The ${base64_decode(secretName)} delete success`)
+                    } else {
+                        console.error(`delete secret thread :: The ${base64_decode(secretName)} delete failed`)
+                    }
+                } catch (error) {
+                    console.error('error _deleteSecret :: ', error)
+                }
+                // remove exist timer
+                remove_delete_task(secretName, appid, DB)
+            }, delayMillisecond, DB, appid, secretName)
+            //add delete task in queue
+            delete_task_queue[key] = {
+                timer,
+                plannedDeleteTime
+            }
+        }
+    } catch (error) {
+        console.error('error deleteSecret :: ', error)
+    }
+}
+
+/**
+ * Initialize the delete thread and call it when the server starts
+ * @param {Object} DB : database control
+ * @returns
+ */
+const _secret_delete_timer = (DB) => {
+    //Execute timer every six hours
+    const timer = setInterval(async () => {
+        try {
+            console.info('in interval')
+            //Find secretName and appid in secret_metadata through planeddeletetime
+            const query_plannedDeleteTime = {
+                selector: {
+                    deleteTime: { "$ne": null }
+                },
+                fields: ['appid', 'secretName', 'plannedDeleteTime']
+            }
+            let plannedDeleteTime_res = await DB.partitionedFind('secret_metadata', query_plannedDeleteTime)
+            if (plannedDeleteTime_res.docs.length > 0) {
+                for (var i = 0; i < plannedDeleteTime_res.docs.length; i++) {
+                    //add deleted tasks by planning deletetime
+                    add_delete_task(DB, plannedDeleteTime_res.docs[i]['appid'],
+                        plannedDeleteTime_res.docs[i]['secretName'],
+                        plannedDeleteTime_res.docs[i]['plannedDeleteTime'])
+                }
+            } else {
+                return
+            }
+        } catch (error) {
+            console.error('error startTimer :: ', error)
+        }
+    }, 6 * 1000 * 60 * 60, DB)
+    return { timer }
+}
+
+module.exports = {
+    _secret_delete_timer,
+    add_delete_task,
+    remove_delete_task
+}
--- a/ehsm_kms_service/ehsm_kms_server.js
+++ b/ehsm_kms_service/ehsm_kms_server.js
@@ -20,6 +20,9 @@ const {
   router,
   GetRouter
 } = require('./router')
+const {
+  _secret_delete_timer
+} = require('./delete_secret_thread')
 
 const app = express()
 app.use(express.json())
@@ -35,6 +38,12 @@ const server = (DB) => {
     console.log('service Initialize exception!')
     process.exit(0)
   }
+
+
+  /**
+   * start secret delete timer
+   */
+  const { timer: secret_delete_timer } = _secret_delete_timer(DB)
 
   /**
    * Clear nonce cache for more than 15 minutes
@@ -65,6 +74,7 @@ const server = (DB) => {
     ehsm_napi.NAPI_Finalize()
     clearInterval(nonce_cache_timer)
     clearInterval(cmk_cache_timer)
+    clearInterval(secret_delete_timer)
     process.exit(0)
   })
 
--- a/ehsm_kms_service/secret_manager_apis.js
+++ b/ehsm_kms_service/secret_manager_apis.js
@@ -19,6 +19,10 @@ const {
     cryptographic_apis,
 } = require('./apis')
 const ehsm_napi = require('./ehsm_napi')
+const {
+    add_delete_task,
+    remove_delete_task
+  } = require('./delete_secret_thread')
 
 /**
  * get a string parameter from payload and base64 if you need
@@ -779,6 +783,7 @@ const deleteSecret = async (res, appid, DB, payload) => {
         if (recoveryPeriod == '' || recoveryPeriod == undefined) {
             recoveryPeriod = DEFAULT_DELETE_RECOVERY_DAYS
         }
+        let plannedDeleteTime = deleteTime + 24 * 60 * 60 * 1000 * recoveryPeriod
 
         //Query and change the secret metadata
         const query_matadata = {
@@ -793,6 +798,7 @@ const deleteSecret = async (res, appid, DB, payload) => {
             if (forceDelete == 'true') {
                 //Force delete secret metadata and secret version data by appId and secretName
                 if (await forceDeleteData(DB, appid, secretName)) {
+                    remove_delete_task(secretName, appid)
                     res.send(_result(200, `The ${base64_decode(secretName)} delete success`))
                     return
                 } else {
@@ -802,8 +808,9 @@ const deleteSecret = async (res, appid, DB, payload) => {
             } else {
                 //update secret metadata
                 secret_metadata_res.docs[0].deleteTime = deleteTime
-                secret_metadata_res.docs[0].plannedDeleteTime = deleteTime + 24 * 60 * 60 * 1000 * recoveryPeriod
+                secret_metadata_res.docs[0].plannedDeleteTime = plannedDeleteTime
                 await DB.insert(secret_metadata_res.docs[0])
+                add_delete_task(DB, appid, secretName, plannedDeleteTime)
                 res.send(_result(200, `The ${base64_decode(secretName)} will be delete after ${recoveryPeriod}d`))
                 return
             }
@@ -966,6 +973,8 @@ const restoreSecret = async (res, appid, DB, payload) => {
             secret_metadata_res.docs[0].deleteTime = null
             secret_metadata_res.docs[0].plannedDeleteTime = null
             await DB.insert(secret_metadata_res.docs[0])
+            //Delete the added thread by secretName
+            remove_delete_task(secretName, appid)
             res.send(_result(200, `The ${base64_decode(secretName)} restore success.`))
             return
         } else {
@@ -991,5 +1000,6 @@ module.exports = {
     describeSecret,
     deleteSecret,
     getSecretValue,
-    restoreSecret
+    restoreSecret,
+    forceDeleteData
 }


ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,131 @@
...
+}
+
+/**
+ * Initialize the delete thread and call it when the server starts
+ * @param {Object} DB : database control
+ * @returns
+ */
+const init_secret_delete_thread = (DB) => {
+    if (init_interval == undefined || init_interval._destroyed) {
+        init_interval = setInterval(_initTimer, 1000 * 60 * 60 * 6, DB)
```
yang8621, line 94: execute per 6h? 
The minimum planned delete time is 1day afterward, right?
Seems 6h is too long.

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,131 @@
...
+        // If the new time is within plus or minus 10 seconds of the original time, keep the original plan unchanged; otherwise, clear the original plan and add again
+        if (exist_timer.hasOwnProperty(secretName)) {
+            if (exist_timer[secretName].plannedDeleteTime + 10000 > plannedDeleteTime && exist_timer[secretName].plannedDeleteTime - 10000 < plannedDeleteTime) {
+                needAdd = false
+            } else {
+                clearTimeout(exist_timer[secretName])
+            }
+        }
+        //If the planned delete time exceeds 12 hours, the thread will not be started
+        if (delayMillisecond > 12 * 1000 * 60 * 60) {
```
yang8621, line 49: we can delete the secret later than the planned time due to the scheduled interval time. But cannot delete it earlier than the time even it's 1s.

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,131 @@
...
+ * @param {String} appid : appid of user
+ * @param {Object} DB : database control
+ * @param {Object} secretName : The name of the secret. eg. 'secretName01'
+ * @param {Object} plannedDeleteTime : planned delete time for secret eg. '1659931644416'
+ * @returns
+ */
+const startDeleteSecretTask = (DB, appid, secretName, plannedDeleteTime) => {
+    try {
+        let delayMillisecond = plannedDeleteTime - new Date().getTime()
+        let needAdd = true
```
yang8621, line 38: what does needAdd mean?  need add what?   Or deleteRequired?

Yidingluo1: it means need to add a deleted task

ehsm_kms_service/delete_secret_thread.js
Yidingluo1: it will not be deleted early until the scheduled time

ehsm_kms_service/delete_secret_thread.js
Yidingluo1: understood，what about execute once an hour

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,137 @@
...
+ * Expose the interface externally, scheduled tasks of delete secret
+ * @param {String} appid : appid of user
+ * @param {Object} DB : database control
+ * @param {Object} secretName : The name of the secret. eg. 'secretName01'
+ * @param {Object} plannedDeleteTime : planned delete time for secret eg. '1659931644416'
+ * @returns
+ */
+const startDeleteSecretTask = (DB, appid, secretName, plannedDeleteTime) => {
+    try {
+        let delayMillisecond = plannedDeleteTime - new Date().getTime()
```
yang8621, line 41: Is it possible that delayMillisecond < 0?
For the case that server is down for several minutes or hours and then gets recovered?

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,137 @@
...
+ * @param {Object} key : key form exist timer
+ * @returns
+ */
+const removeTimer = (key, DB) => {
+    try {
+        //Delete thread when thread exists
+        if (exist_timer[key] != undefined ) {
+            clearTimeout(exist_timer[key].timer)
+        }
+        delete exist_timer[key];
```
yang8621, line 23: @line23, How about exist_timer[key] id undefined?

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,137 @@
...
+// Record timer waiting to run
+var exist_timer = {}
+
+/**
+ * remove a timer from exist timer by key
+ * @param {Object} DB : database control
+ * @param {Object} key : key form exist timer
+ * @returns
+ */
+const removeTimer = (key, DB) => {
```
yang8621, line 17: Do you think both removeTimer and exist_timer requires to bind **1. secret name** together with **2. appid**?

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,137 @@
+const {
+    base64_decode
+} = require('./function')
+
+// Example of setInterval
+var init_interval
+
+// Record timer waiting to run
+var exist_timer = {}
```
yang8621, line 9: rename to timer_queue

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,137 @@
...
+    try {
+        let delayMillisecond = plannedDeleteTime - new Date().getTime()
+        let needAdd = true
+        // process exist timer, 
+        // If the new time is within plus or minus 10 seconds of the original time, keep the original plan unchanged; otherwise, clear the original plan and add again
+        if (exist_timer.hasOwnProperty(secretName)) {
+            if (exist_timer[secretName].plannedDeleteTime + 10000 > plannedDeleteTime && exist_timer[secretName].plannedDeleteTime - 10000 < plannedDeleteTime) {
+                needAdd = false
+            } else {
+                clearTimeout(exist_timer[secretName].timer)
```
yang8621, line 49: not sure why need to clearTimeout here.

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * Initialize the delete thread and call it when the server starts
+ * @param {Object} DB : database control
+ * @returns
+ */
+const init_secret_delete_thread = (DB) => {
+    if (init_interval == undefined || init_interval._destroyed) {
+        //Execute _initTimer every six hours
+        init_interval = setInterval(_initTimer, 1000 * 60 * 60 * 6, DB)
```
yang8621, line 99: setInterval 6h, will run the thread @0h, 6h, 12h... OR @6h, 12h,18h ... ?  Which one?

ehsm_kms_service/delete_secret_thread.js
Yidingluo1: if the value is less than 0, it will be directly deleted during the first run after the server is started

ehsm_kms_service/delete_secret_thread.js
Yidingluo1: during restore, if the planned delete time is more than 12 hours, the timer will be undefined, that is, no deletion thread will be added

ehsm_kms_service/delete_secret_thread.js
Yidingluo1: ok, it will be changed

ehsm_kms_service/delete_secret_thread.js
yang8621: So why check @line 20? 
I'm curious why check undefined only for line21 but not line 23.

ehsm_kms_service/delete_secret_thread.js
yang8621: where the code?

ehsm_kms_service/delete_secret_thread.js
Yidingluo1: when add a delete thread repeatedly, the latter will overwrite the previous one. Here, need to delete the previous thread

ehsm_kms_service/delete_secret_thread.js
Yidingluo1: 6H, 12h,18h

ehsm_kms_service/delete_secret_thread.js
Yidingluo1: @setTimeout() if the parameter is negative, it will run directly

ehsm_kms_service/delete_secret_thread.js
Yidingluo1: line 23 is delete whole exist_timer, line 21 just deletes the timer created by setTimeout()

ehsm_kms_service/delete_secret_thread.js
Yidingluo1: I think just need the secretName as the key enough

ehsm_kms_service/delete_secret_thread.js
yang8621: Thanks.  It depends on the API SetTimeout(). 
But when I look up the API from here: https://nodejs.org/api/timers.html#settimeoutcallback-delay-args
I found a new issue: 
_**When delay is larger than 2147483647 or less than 1, the delay will be set to 1. Non-integer delays are truncated to an integer.**_

2147483647 / 1000 / 60 / 60 / 24 = 24.85 day
But what is the maximum we would support?

Yidingluo1: The maximum is 365 days

yang8621: Thanks     That's fine.

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,137 @@
...
+ * @param {Object} key : key form exist timer
+ * @returns
+ */
+const removeTimer = (key, DB) => {
+    try {
+        //Delete thread when thread exists
+        if (timer_queue[key] != undefined ) {
+            clearTimeout(timer_queue[key].timer)
+        }
+        delete timer_queue[key];
```
yang8621, line 23: If timer_queue[key] is undefined, will line 23 crash?

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,137 @@
...
+        // An insurance mechanism to ensure that thread init_interval is active
+        init_secret_delete_thread(DB)
```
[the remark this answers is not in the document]
Yidingluo1: this is a protection mechanism, if the_InitTimer in @line99 stopped unexpectedly, it will run again here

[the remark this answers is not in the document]
Yidingluo1: this is a protection mechanism, if the_InitTimer in LINE99 stopped unexpectedly, it will run again here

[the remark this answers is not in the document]
yang8621: > this is a protection mechanism, if the_InitTimer in LINE99 stopped unexpectedly, it will run again here

If not stopped, what's impact if call again?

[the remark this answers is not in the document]
Yidingluo1: no impact, because of line97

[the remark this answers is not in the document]
yang8621: That's fine.  Thanks.

yang8621 (approved):

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,137 @@
...
+            }
+        }
+        //If the planned delete time exceeds 12 hours, the thread will not be started
+        if (delayMillisecond > 12 * 1000 * 60 * 60) {
+            needAdd = false
+        }
+        if (needAdd) {
+            //Force delete according to planned deletion time
+            let timer = setTimeout(_deleteSecret, delayMillisecond, DB, appid, secretName)
+            timer_queue[secretName] = {
```
HawieLong, line 59: what would happen if two users have the same secret name and want to delete their secret simultaneously? the timer queue[secretName] can support this scenario?

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,137 @@
...
+        if (await forceDeleteData(DB, appid, secretName)) {
+            console.info(`delete secret thread :: The ${base64_decode(secretName)} delete success`)
+        } else {
+            console.error(`delete secret thread :: The ${base64_decode(secretName)} delete failed`)
+        }
+    } catch (error) {
+        console.error('error _deleteSecret :: ', error)
+    }
+    // remove exist timer
+    removeTimer(secretName, DB)
```
HawieLong, line 88: Remove timer only use secretName, can this function remove other's secret which has the same secret name?

Yidingluo1: No, the secretname is searched according to the deletetime. only the secretname with the deletion time will be deleted

yang8621: How about two appid have two same secretname and in very close planned delete time

ehsm_kms_service/delete_secret_thread.js
Yidingluo1: This is a problem. The timer queue[secretName] will be changed to secretname + '_'+ appid

ehsm_kms_service/delete_secret_thread.js
Yidingluo1: according to the planned delete time, the earlier one will be deleted first

ehsm_kms_service/ehsm_kms_server.js
```diff
@@ -36,6 +39,12 @@ const server = (DB) => {
     process.exit(0)
   }
 
+
+  /**
+   * start secret delete thread
+   */
+  init_secret_delete_thread(DB)
```
syan10, line 46: why not follow the line#51 to name it as a xxx_timer()?

Yidingluo1: what about _secret_delete_timer?

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,138 @@
...
+        if (await forceDeleteData(DB, appid, secretName)) {
+            console.info(`delete secret thread :: The ${base64_decode(secretName)} delete success`)
+        } else {
+            console.error(`delete secret thread :: The ${base64_decode(secretName)} delete failed`)
+        }
+    } catch (error) {
+        console.error('error _deleteSecret :: ', error)
+    }
+    // remove exist timer
+    removeTimer(secretName, appid, DB)
```
syan10, line 89: does it mean after the secret be deleted, the timer also will be removed？

seems the other existing timer (e,g, nonce_timer, cmk_timer) only will be remove after the service stopped.  can you explain why?

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,138 @@
+const {
+    base64_decode
+} = require('./function')
+
+// Example of setInterval
```
syan10, line 5: from the comment, i can't get what does it mean?

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,138 @@
...
+var init_interval
+
+// Record timer waiting to run
```
syan10, line 8: same here

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,138 @@
...
+ * @param {String} appid : appid of user
+ * @param {Object} DB : database control
+ * @param {Object} secretName : The name of the secret. eg. 'secretName01'
+ * @param {Object} plannedDeleteTime : planned delete time for secret eg. '1659931644416'
+ * @returns
+ */
+const startDeleteSecretTask = (DB, appid, secretName, plannedDeleteTime) => {
+    try {
+        let delayMillisecond = plannedDeleteTime - new Date().getTime()
+        let needAdd = true
```
syan10, line 42: needAdd means?

syan10: please use meaningful variables.

ehsm_kms_service/delete_secret_thread.js
syan10: seems you can combine the _initTimer and init_secret_delete_thread into one function, which can be named as _secret_delete_timer

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * Expose the interface externally, scheduled tasks of delete secret
+ * @param {String} appid : appid of user
+ * @param {Object} DB : database control
+ * @param {Object} secretName : The name of the secret. eg. 'secretName01'
+ * @param {Object} plannedDeleteTime : planned delete time for secret eg. '1659931644416'
+ * @returns
+ */
+const startDeleteSecretTask = (DB, appid, secretName, plannedDeleteTime) => {
+    try {
```
syan10, line 40: based on my understanding, seems this function is used to add a new delete task into the delete queue with the secret name.
if yes, suggest to rename as add_delete_task(),  and the corresponding function of removeTimer renamed as remove_delete_task().

and the timer_queue renamed as delete_task_queue

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,138 @@
...
+        // An insurance mechanism to ensure that thread init_interval is active
+        init_secret_delete_thread(DB)
+    } catch (error) {
+        console.error('error removeTimer :: ', error)
+    }
+}
+
+/**
+ * Expose the interface externally, scheduled tasks of delete secret
+ * @param {String} appid : appid of user
```
syan10, line 33: also please update the comment to give a more meaningful description.

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,138 @@
+const {
+    base64_decode
+} = require('./function')
+
+// Example of setInterval
+var init_interval
```
syan10, line 6: see the below comments.

ehsm_kms_service/delete_secret_thread.js
Yidingluo1: need add a delete secret timer

Yidingluo1:
@syan10 all comments changed

ehsm_kms_service/ehsm_kms_server.js
```diff
@@ -36,6 +39,12 @@ const server = (DB) => {
     process.exit(0)
   }
 
+
+  /**
+   * start secret delete thread
+   */
+  _secret_delete_timer(DB)
```
syan10, line 46: do you need to clear it after the service is down?

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,123 @@
...
+ * @returns
+ */
+const remove_delete_task = (secretName, appid, DB) => {
+    try {
+        //Delete thread when thread exists
+        if (delete_task_queue[secretName + '_' + appid] != undefined ) {
+            clearTimeout(delete_task_queue[secretName + '_' + appid].timer)
+        }
+        delete delete_task_queue[secretName + '_' + appid];
+        // An insurance mechanism to ensure that thread interval_time_instance is active
```
syan10, line 24: does it actually need? if it have issue, how to you consider the other timer? they also should have the same issue..

Yidingluo1: ok, I will delete it

syan10:
add secret delete timer

it will scan the database every 6 hours and delete items that have reached the deletion time.

passed the python test

Yidingluo1:
> 

commit changed

Yidingluo1:
> add secret delete timer
> 
> it will scan the database every 6 hours and delete items that have reached the deletion time.
> 
> passed the python test Signed-off-by: yidingluo1 [luo.yd@neusoft.com](mailto:luo.yd@neusoft.com)

commit changed

syan10 (approved):

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,118 @@
+const {
+    base64_decode
+} = require('./function')
+
+// Set a queue to record the waiting time
+var delete_task_queue = {}
```
HawieLong, line 6: What would happen if I use forcedelete API deleted a secret already in delete_task_queue? The secret would be deleted twice?

Yidingluo1: forcibly deleted content will be deleted directly when running secret delete api, and will not be added to delete_task_queue

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,118 @@
...
+ * @returns
+ */
+const add_delete_task = (DB, appid, secretName, plannedDeleteTime) => {
+    try {
+        let delayMillisecond = plannedDeleteTime - new Date().getTime()
+        let needAdd = true
+        let key = secretName + '_' + appid
+        // process exist timer, 
+        // If the new time is within plus or minus 10 seconds of the original time, keep the original plan unchanged; otherwise, clear the original plan and add again
+        if (delete_task_queue.hasOwnProperty(key)) {
```
HawieLong, line 41: If a user pushes the delete time 8 seconds later and wants to restore the secret in the last 5 seconds. This would cause this secret cannot be found. Is this acceptable?

Yidingluo1: Scheduled deletion time is at least 1 day，8 seconds is too short

ehsm_kms_service/delete_secret_thread.js
HawieLong: I mean the secret is planned to be deleted, for example, 8 hours, at first.
But I use force delete when there are 2 hours left and has been added in delete_task_queue.
What would happen?

Yidingluo1: This is a problem, it will be modified to after force deletion, delete this delete_task

ehsm_kms_service/delete_secret_thread.js
HawieLong: I schedule to delete recoveryPeriod=2 days on 2022.1.1 13:00:00
then I schedule to delete recoveryPeriod=1 on 2022.1.2 13:00:08
that will get a 8 seconds later.

ehsm_kms_service/delete_secret_thread.js
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+/**
+ * Initialize the delete thread and call it when the server starts
+ * @param {Object} DB : database control
+ * @returns
+ */
+const _secret_delete_timer = (DB) => {
+    //Execute _initTimer every six hours
```
HawieLong, line 85: change this note to new name

PR #158, opened 2022-08-12
init framework for sgx-ssl

whh39:


core/App/ehsm_napi.cpp
```diff
@@ -94,8 +94,11 @@ void NAPI_Finalize(){
         }
     }
 */
-char* NAPI_CreateKey(const uint32_t keyspec, const uint32_t origin)
+char* NAPI_CreateKey(const char* paramJson)
```
syan10, line 97: suggest to define the interfaces as:
NAPI_CreateKey(const char* CreateKeyRequest,  char* CreateKeyResponse)
NAPI_Encrypt(const char* EncryptRequest,  char* EncryptResponse)
...

btw, i'm not sure if the parameter in json format is a formal way to do the cross-language program.

wanghouqi: need try NAPI_CreateKey(const char* CreateKeyRequest, char* CreateKeyResponse)

PR #160, opened 2022-08-19
Sgx ssl framework

whh39:


core/Enclave/enclave_hsm.edl
```diff
@@ -67,49 +67,42 @@ enclave {
...
-                                            [in, size=ciphertext_len] const uint8_t *ciphertext, uint32_t ciphertext_len,
-                                            [out, size=plaintext_len] uint8_t *plaintext, uint32_t plaintext_len,
-                                            [out] uint32_t *req_plaintext_len);
+        public sgx_status_t enclave_create_aes_key([out, size=cmk_blob_size] uint8_t *cmk_blob, uint32_t cmk_blob_size);
+        
+        public sgx_status_t enclave_create_rsa_key([out, size=cmk_blob_size] uint8_t *cmk_blob, uint32_t cmk_blob_size);
+
+        //public sgx_status_t enclave_create_ec_key(/* param */);
+
+        //public sgx_status_t enclave_create_hmac_key(/* param */);
```
syan10, line 76: you need to be notice that, the number of ecall is as smaller as good.

so you better to combine the similar function into one. such as:
createkey()
encrypt()
decrypt()
sign()
verify()
....

include/datatypes.h
```diff
@@ -134,7 +108,104 @@ typedef struct _session_id_tracker_t
...
+    uint32_t purpose;
+    uint32_t apiversion;
+    uint8_t  descrption[16];
+    uint8_t  createdate[8];
+} ehsm_keymetadata_t;
+
+typedef struct {
+    ehsm_keymetadata_t  metadata;       //common key properties
+    uint32_t            keybloblen;     
+    uint8_t             *keyblob;       //private key properties
```
syan10, line 127: it's not only for private key

include/datatypes.h
```diff
@@ -134,7 +108,104 @@ typedef struct _session_id_tracker_t
     uint32_t          session_id;
 } session_id_tracker_t;
 
-#pragma pack(pop)
+//sgx-ssl framework
+
+typedef struct {
+    uint32_t keyspec;
+    uint32_t digest_mdoe;
+    uint32_t padding_mdoe;
```
syan10, line 116: typo, mdoe->mode

include/datatypes.h
```diff
@@ -134,7 +108,104 @@ typedef struct _session_id_tracker_t
...
+    uint32_t origin;
+    uint32_t purpose;
+    uint32_t apiversion;
+    uint8_t  descrption[16];
```
syan10, line 120: typo, description

include/datatypes.h
syan10: the modes are not common parameters, you can put them into a separate struct as "addition", such as defined them as:
typedef struct {
  uint32_t keyspec;
  uint32_t origin;
  uint32_t purpose;
  uint32_t apiversion;
  uint8_t  descrption[16];
  uint8_t createdate[8];
  struct addition addition;
}

you can try to find a better name.

core/App/ehsm_provider.cpp
```diff
@@ -130,7 +136,187 @@ void Finalize()
...
+
+/**
+ * @brief Initialize the variables needed to generate the aes_gcm key
+ * 
+ * @param sgxStatus result for enclave function
+ * @param cmk storage key information
+ * @param paramJson Pass in the key parameter in the form of JSON string
+ * @return sgx_status_t 
+ */
+sgx_status_t create_aes_key(sgx_status_t* sgxStatus, ehsm_keyblob_t* cmk, JsonObj paramJson)
```
syan10, line 165: does it need to pass the paramJson?

PR #163, opened 2022-08-24
add support for rsa createKey/encrypy/decrypt with sgx-ssl api

whh39:
it can create rsa key and length is 2048 or 3072
padding mode supports RSA_PKCS1_PADDING and RSA_PKCS1_OAEP_PADDING

add test case for rsa_2048/3072_encrypt/decrypt

test: passed the unit test

Signed-off-by: wanghaiheng <wanghaiheng@neusoft.com>

### DIFF
--- a/core/App/ehsm_napi.cpp
+++ b/core/App/ehsm_napi.cpp
@@ -44,12 +44,13 @@
 #include "ehsm_marshal.h"
 #include "auto_version.h"
 
+#include "openssl/rsa.h"
+
 using namespace std;
 using namespace EHsmProvider;
 
 
 extern "C" {
-
 /*
 create the enclave
 @return
@@ -135,7 +136,15 @@ char* NAPI_CreateKey(const char* paramJson)
     master_key.metadata.keyspec = paramJsonObj.readData_uint16("keyspec");
     master_key.metadata.origin = paramJsonObj.readData_uint16("origin");
     master_key.metadata.purpose = paramJsonObj.readData_uint16("purpose");
-    master_key.keybloblen = 0;    
+    master_key.metadata.padding_mode = paramJsonObj.readData_uint16("padding_mode");
+    master_key.metadata.digest_mode = paramJsonObj.readData_uint16("digest_mode");
+    master_key.keybloblen = 0;
+
+    if (master_key.metadata.padding_mode == RSA_NO_PADDING) {
+        retJsonObj.setCode(retJsonObj.CODE_FAILED);
+        retJsonObj.setMessage("NO padding is unsafe.");
+        goto out;
+    } 
     
     // get keybloblen
     ret = CreateKey(&master_key);
@@ -247,6 +256,7 @@ char* NAPI_Encrypt(const char* paramJson)
     int plaintext_len = plaintext_str.size();
     int aad_len = aad_str.size();
 
+    printf("cmk_len=%d\n", cmk_len);
     if(cmk_len == 0 || cmk_len > EH_CMK_MAX_SIZE){
         retJsonObj.setCode(retJsonObj.CODE_BAD_REQUEST);
         retJsonObj.setMessage("The cmk's length is invalid.");
@@ -1007,10 +1017,9 @@ char* NAPI_AsymmetricEncrypt(const char* paramJson)
     switch (cmk.metadata.keyspec)
     {
         case EH_RSA_2048:
-            // TODO
-            break;
         case EH_RSA_3072:
-            plaintext_maxLen = RSA_OAEP_3072_SHA_256_MAX_ENCRYPTION_SIZE;
+        case EH_RSA_4096:
+            plaintext_maxLen = ehsm_get_rsa_max_encryption_size(cmk.metadata.keyspec, cmk.metadata.padding_mode);
             break;
         case EH_EC_SM2:
             // TODO
@@ -1047,6 +1056,7 @@ char* NAPI_AsymmetricEncrypt(const char* paramJson)
 
     ret = AsymmetricEncrypt(&cmk, &plaint_data, &cipher_data);
     if (ret != EH_OK) {
+
         retJsonObj.setCode(retJsonObj.CODE_FAILED);
         retJsonObj.setMessage("Server exception.");
         goto out;
@@ -1135,10 +1145,9 @@ char* NAPI_AsymmetricDecrypt(const char* paramJson)
     switch (cmk.metadata.keyspec)
     {
         case EH_RSA_2048:
-            // TODO
-            break;
         case EH_RSA_3072:
-            ciphertext_maxLen = RSA_OAEP_3072_CIPHER_LENGTH;
+        case EH_RSA_4096:
+            ciphertext_maxLen = ehsm_get_rsa_cipher_len(cmk.metadata.keyspec);
             break;
         case EH_EC_SM2:
             // TODO
@@ -1179,7 +1188,7 @@ char* NAPI_AsymmetricDecrypt(const char* paramJson)
         retJsonObj.setMessage("Server exception.");
         goto out;
     }
-    
+
     plaintext_base64 = base64_encode(plaint_data.data, plaint_data.datalen);
     if(plaintext_base64.size() > 0){
         retJsonObj.addData_string("plaintext", plaintext_base64);
@@ -1276,11 +1285,12 @@ char* NAPI_ExportDataKey(const char* paramJson)
         retJsonObj.setMessage("The aad's length is invalid.");
         goto out;
     }
-    if(olddatakey_len == 0 || olddatakey_len > RSA_OAEP_3072_SHA_256_MAX_ENCRYPTION_SIZE){
-        retJsonObj.setCode(retJsonObj.CODE_BAD_REQUEST);
-        retJsonObj.setMessage("The olddatakey's length is invalid.");
-        goto out;
-    }
+    // close there code temporary: RSA_OAEP_3072_SHA_256_MAX_ENCRYPTION_SIZE is deprecated
+    // if(olddatakey_len == 0 || olddatakey_len > RSA_OAEP_3072_SHA_256_MAX_ENCRYPTION_SIZE){
+    //     retJsonObj.setCode(retJsonObj.CODE_BAD_REQUEST);
+    //     retJsonObj.setMessage("The olddatakey's length is invalid.");
+    //     goto out;
+    // }
 
     ret = ehsm_deserialize_cmk(&cmk, (const uint8_t*)cmk_str.data(), cmk_len);
     if (ret != EH_OK) {
--- a/core/App/ehsm_provider.cpp
+++ b/core/App/ehsm_provider.cpp
@@ -149,23 +149,6 @@ void Finalize()
     sgx_destroy_enclave(g_enclave_id);
 }
 
-static uint32_t sgx_calc_gcm_data_size(const uint32_t aad_size, const uint32_t plaintext_size)
-{
-    if (aad_size > UINT32_MAX - sizeof(aes_gcm_key_data_t))
-        return UINT32_MAX;
-
-    if (plaintext_size > UINT32_MAX - sizeof(aes_gcm_key_data_t))
-        return UINT32_MAX;
-
-    if (aad_size > UINT32_MAX - plaintext_size)
-        return UINT32_MAX;
-
-    if (sizeof(aes_gcm_key_data_t) > UINT32_MAX - plaintext_size - aad_size)
-        return UINT32_MAX;
-
-    return (aad_size + plaintext_size + sizeof(aes_gcm_key_data_t));
-}
-
 /**
  * @brief Initialize the variables needed to generate the aes_gcm key
  * 
@@ -176,64 +159,7 @@ static uint32_t sgx_calc_gcm_data_size(const uint32_t aad_size, const uint32_t p
  */
 sgx_status_t create_aes_key(sgx_status_t* sgxStatus, ehsm_keyblob_t* cmk, JsonObj paramJson)
 {
-    uint32_t key_size;
-    if (cmk->keybloblen == 0)
-    {
-        //TODO: compute keybloblen
-        
-        switch (cmk->metadata.keyspec)
-        {
-            case EH_AES_GCM_128:
-                key_size = 16;
-                break;
-            case EH_AES_GCM_192:
-                key_size = 24;
-                break;
-            case EH_AES_GCM_256:
-                key_size = 32;
-                break;
-            default:
-                return SGX_ERROR_INVALID_PARAMETER;
-        }
-        cmk->keybloblen = sgx_calc_gcm_data_size(0, key_size);
-        return SGX_SUCCESS;
-    }
     
-    return enclave_create_aes_key(g_enclave_id, sgxStatus, cmk->keyblob, cmk->keybloblen);
-}
-
-/**
- * @brief Initialize the variables needed to generate the rsa key
- * 
- * @param sgxStatus result for enclave function
- * @param cmk storage key information
- * @param paramJson Pass in the key parameter in the form of JSON string
- * @return sgx_status_t 
- */
-sgx_status_t create_rsa_key(sgx_status_t* sgxStatus, ehsm_keyblob_t* cmk, JsonObj paramJson)
-{
-    if (cmk->keybloblen == 0)
-    {
-        //TODO: compute keybloblen
-        cmk->keybloblen = sgx_calc_gcm_data_size(0, sizeof(rsa_key_data_t));
-
-        return SGX_SUCCESS;
-    }
-    uint32_t key_size;
-    switch (cmk->metadata.keyspec)
-    {
-        case EH_RSA_2048:
-            key_size = 256;
-            break;
-        case EH_RSA_3072:
-            key_size = 384;
-            break;
-        default:
-            return SGX_ERROR_INVALID_PARAMETER;
-    }
-    //TODO: initialize variable
-
-    // return enclave_create_rsa_key(g_enclave_id, sgxStatus, cmk->keyblob);
 }
 
 /**
@@ -348,13 +274,14 @@ ehsm_status_t CreateKey(ehsm_keyblob_t *cmk)
             break;
         case EH_RSA_2048:
         case EH_RSA_3072:
+        case EH_RSA_4096:
             if (cmk->keybloblen == 0)
                 ret = enclave_create_rsa_key(g_enclave_id,
                                              &sgxStatus,
                                              NULL,
                                              0,
                                              &(cmk->keybloblen),
-                                             NULL,
+                                             cmk->metadata.keyspec,
                                              NULL);
             else
                 ret = enclave_create_rsa_key(g_enclave_id,
@@ -482,19 +409,20 @@ ehsm_status_t AsymmetricEncrypt(ehsm_keyblob_t *cmk,
 
     switch(cmk->metadata.keyspec) {
         case EH_RSA_2048:
-            break;
         case EH_RSA_3072:
-            if(plaintext->datalen > RSA_OAEP_3072_SHA_256_MAX_ENCRYPTION_SIZE) {
-                printf("Error data len(%d) for rsa encryption, max is 318.\n", plaintext->datalen);
+        case EH_RSA_4096:
+            if(plaintext->datalen > ehsm_get_rsa_max_encryption_size(cmk->metadata.keyspec, cmk->metadata.padding_mode)) {
+                printf("Error data len(%d) for rsa encryption, max is (%d).\n", 
+                       plaintext->datalen, ehsm_get_rsa_max_encryption_size(cmk->metadata.keyspec, cmk->metadata.padding_mode));
                 return EH_ARGUMENTS_BAD;
             }
             /* calculate the ciphertext length  */
             if (ciphertext->datalen == 0) {
-                ciphertext->datalen = RSA_OAEP_3072_CIPHER_LENGTH;
+                ciphertext->datalen = ehsm_get_rsa_cipher_len(cmk->metadata.keyspec);
                 return EH_OK;
             }
             /* check if the datalen is valid */
-            if (ciphertext->data == NULL || ciphertext->datalen != RSA_OAEP_3072_CIPHER_LENGTH)
+            if (ciphertext->data == NULL || ciphertext->datalen != ehsm_get_rsa_cipher_len(cmk->metadata.keyspec))
                 return EH_ARGUMENTS_BAD;
             ret = enclave_rsa_encrypt(g_enclave_id,
                                       &sgxStatus,
@@ -514,8 +442,9 @@ ehsm_status_t AsymmetricEncrypt(ehsm_keyblob_t *cmk,
             return EH_KEYSPEC_INVALID;
     }
 
-    if (ret != SGX_SUCCESS || sgxStatus != SGX_SUCCESS)
+    if (ret != SGX_SUCCESS || sgxStatus != SGX_SUCCESS) {
         return EH_FUNCTION_FAILED;
+    }
     else
         return EH_OK;
 }
@@ -595,22 +524,23 @@ ehsm_status_t AsymmetricDecrypt(ehsm_keyblob_t *cmk,
     switch(cmk->metadata.keyspec) {
         case EH_RSA_2048:
         case EH_RSA_3072:
+        case EH_RSA_4096:
             if (ciphertext->data == NULL || ciphertext->datalen == 0 ||
-                ciphertext->datalen > RSA_OAEP_3072_CIPHER_LENGTH) {
+                ciphertext->datalen > ehsm_get_rsa_cipher_len(cmk->metadata.keyspec)) {
                 return EH_ARGUMENTS_BAD;
             }
             /* calculate the ciphertext length */
             if (plaintext->datalen == 0) {
                 ret = enclave_rsa_decrypt(g_enclave_id,
-                                  &sgxStatus,
-                                  cmk->keyblob,
-                                  cmk->keybloblen,
-                                  ciphertext->data,
-                                  ciphertext->datalen,
-                                  NULL, 0,
-                                  &(plaintext->datalen),
-                                  cmk->metadata.keyspec,
-                                      cmk->metadata.padding_mode);
+                                          &sgxStatus,
+                                          cmk->keyblob,
+                                          cmk->keybloblen,
+                                          ciphertext->data,
+                                          ciphertext->datalen,
+                                          NULL, 0,
+                                          &(plaintext->datalen),
+                                          cmk->metadata.keyspec,
+                                          cmk->metadata.padding_mode);
                 return EH_OK;
             }
             /* check if the datalen is valid */
@@ -627,6 +557,7 @@ ehsm_status_t AsymmetricDecrypt(ehsm_keyblob_t *cmk,
                                       NULL,
                                       cmk->metadata.keyspec,
                                       cmk->metadata.padding_mode);
+            break;
         case EH_EC_SM2:
             sm2_decrypt(/* param */);
             break;
--- a/core/App/ehsm_provider.h
+++ b/core/App/ehsm_provider.h
@@ -44,28 +44,12 @@
 
 #define ENCLAVE_PATH "libenclave-ehsm-core.signed.so"
 
-#define EH_ENCRYPT_MAX_SIZE (6*1024)
-
 #define EH_AAD_MAX_SIZE (8*1024)
 
 #define EH_QUOTE_MAX_SIZE (8*1024)
 
-#define EH_CMK_MAX_SIZE (4*1024)
+#define EH_CMK_MAX_SIZE (8*1024)
 
-#define EH_DATA_KEY_MAX_SIZE 1024
-
-#define EH_AES_GCM_IV_SIZE  12
-#define EH_AES_GCM_MAC_SIZE 16
-
-#define RSA_OAEP_2048_SHA_256_MAX_ENCRYPTION_SIZE       190
-// #define RSA_2048_OAEP_SHA_1_MAX_ENCRYPTION_SIZE       214
-
-#define RSA_OAEP_3072_SHA_256_MAX_ENCRYPTION_SIZE       318
-// #define RSA_3072_OAEP_SHA_1_MAX_ENCRYPTION_SIZE       342
-
-#define SM2PKE_MAX_ENCRYPTION_SIZE                      6047
-
-#define RSA_OAEP_3072_CIPHER_LENGTH       384
 #define RSA_OAEP_3072_SIGNATURE_SIZE      384
 #define RSA_OAEP_3072_DIGEST_SIZE         256
 
--- a/core/Enclave/enclave_hsm.cpp
+++ b/core/Enclave/enclave_hsm.cpp
@@ -36,6 +36,7 @@
 #include <stdio.h>
 #include <stdbool.h>
 #include <mbusafecrt.h>
+#include <cstring>
 
 #include "sgx_report.h"
 #include "sgx_utils.h"
@@ -49,34 +50,20 @@
 
 using namespace std;
 
-#define SGX_AES_KEY_SIZE 16
-
 #define SGX_DOMAIN_KEY_SIZE     16
 
-#define RSA_OAEP_3072_MOD_SIZE      384
-#define RSA_OAEP_3072_EXP_SIZE      4
-
-#define EH_ENCRYPT_MAX_SIZE (6*1024)
-
-#define EH_DATA_KEY_MAX_SIZE 1024
-
-#define EH_AES_GCM_IV_SIZE  12
-#define EH_AES_GCM_MAC_SIZE 16
-
-#define RSA_OAEP_2048_SHA_256_MAX_ENCRYPTION_SIZE       190
-//#define RSA_2048_OAEP_SHA_1_MAX_ENCRYPTION_SIZE       214
-
-#define RSA_OAEP_3072_SHA_256_MAX_ENCRYPTION_SIZE       318
-//#define RSA_3072_OAEP_SHA_1_MAX_ENCRYPTION_SIZE       342
-
-#define SM2PKE_MAX_ENCRYPTION_SIZE                      6047
-
-#define RSA_OAEP_3072_CIPHER_LENGTH       384
-#define RSA_OAEP_3072_SIGNATURE_SIZE      384
-
-#define RSA_3072_KEY_BITS   3072
 #define RSA_2048_KEY_BITS   2048
+#define RSA_3072_KEY_BITS   3072
+#define RSA_4096_KEY_BITS   4096
 
+#define RSA_2048_PUBLIC_KEY_PEM_SIZE    426
+#define RSA_2048_PRIVATE_KEY_PEM_SIZE    1679
+
+#define RSA_3072_PUBLIC_KEY_PEM_SIZE    601
+#define RSA_3072_PRIVATE_KEY_PEM_SIZE    2459
+
+#define RSA_4096_PUBLIC_KEY_PEM_SIZE    775
+#define RSA_4096_PRIVATE_KEY_PEM_SIZE    3243
 
 // Used to store the secret passed by the SP in the sample code.
 sgx_aes_gcm_128bit_key_t g_domain_key = {0};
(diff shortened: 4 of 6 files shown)


core/App/ehsm_provider.cpp
```diff
@@ -149,91 +149,53 @@ void Finalize()
...
+}
 
-    return (aad_size + plaintext_size + sizeof(aes_gcm_key_data_t));
+static uint32_t sgx_get_rsa_padding_size(uint8_t padding_mode)
+{
+    switch (padding_mode) {
+        case RSA_PKCS1_OAEP_PADDING:
+            return 41;
+        case RSA_PKCS1_PADDING:
+            return 11;
```
syan10, line 170: where did you get these value?

core/App/ehsm_provider.cpp
```diff
@@ -149,91 +149,53 @@ void Finalize()
...
-                key_size = 32;
-                break;
-            default:
-                return SGX_ERROR_INVALID_PARAMETER;
-        }
-        cmk->keybloblen = sgx_calc_gcm_data_size(0, key_size);
-        return SGX_SUCCESS;
+    switch (cmk->metadata.keyspec) {
+        case EH_RSA_2048:
+            return RSA_OAEP_2048_CIPHER_LENGTH - sgx_get_rsa_padding_size(cmk->metadata.padding_mode);
```
syan10, line 180: seems you already defined a function: sgx_get_rsa_cipher_len(), so does it need to use the switch..case here?

core/App/ehsm_provider.cpp
```diff
@@ -149,91 +149,53 @@ void Finalize()
...
-    if (aad_size > UINT32_MAX - plaintext_size)
-        return UINT32_MAX;
-
-    if (sizeof(aes_gcm_key_data_t) > UINT32_MAX - plaintext_size - aad_size)
-        return UINT32_MAX;
+    switch (keyspec) {
+        case EH_RSA_2048:
+            return RSA_OAEP_2048_CIPHER_LENGTH;
+        case EH_RSA_3072:
+            return RSA_OAEP_3072_CIPHER_LENGTH;
```
syan10, line 158: 4096?

core/App/ehsm_provider.cpp
```diff
@@ -482,19 +444,18 @@ ehsm_status_t AsymmetricEncrypt(ehsm_keyblob_t *cmk,
 
     switch(cmk->metadata.keyspec) {
         case EH_RSA_2048:
-            break;
         case EH_RSA_3072:
-            if(plaintext->datalen > RSA_OAEP_3072_SHA_256_MAX_ENCRYPTION_SIZE) {
-                printf("Error data len(%d) for rsa encryption, max is 318.\n", plaintext->datalen);
+            if(plaintext->datalen > sgx_get_rsa_max_encryption_size(cmk)) {
```
syan10, line 448: same logic for both 2048 and 3072?

core/Enclave/enclave_hsm.cpp
```diff
@@ -281,7 +302,19 @@ sgx_status_t enclave_create_rsa_key(uint8_t *cmk_blob,
     sgx_status_t ret = SGX_ERROR_UNEXPECTED;
 
     // get keybloblen
-    uint32_t real_keyblob_size = sgx_calc_gcm_data_size(0, sizeof(rsa_key_data_t));
+    int32_t key_size;
+    switch (keyspec)
+    {
+        case EH_RSA_2048:
+            key_size = 2105;
```
syan10, line 309: do not use hardcode!
it's a MUST requirement for the following patches...

syan10: and also need to describe the details how you get/set this values..

core/Enclave/enclave_hsm.cpp
```diff
@@ -294,43 +327,92 @@ sgx_status_t enclave_create_rsa_key(uint8_t *cmk_blob,
         printf("ecall create_rsa_key cmk_keyblob_size:%lu < key_blob_size:%d.\n", cmk_blob_size, real_keyblob_size);
         return SGX_ERROR_INVALID_PARAMETER;
     }
-        rsa_key_data_t *keyblob = (rsa_key_data_t *)cmk_blob;
+    if (keyspec == NULL || padding_mode == NULL)
+        return SGX_ERROR_INVALID_PARAMETER;
+
+    RSA     *rsa            = NULL;
+    BIGNUM  *bne            = NULL;
```
syan10, line 334: the name is not good!  please following the sgx-ssl rules.

core/Enclave/enclave_hsm.cpp
```diff
@@ -294,43 +327,92 @@ sgx_status_t enclave_create_rsa_key(uint8_t *cmk_blob,
...
+            ret = SGX_ERROR_OUT_OF_MEMORY;
+            break;
+        }
+
+        if(!BN_set_word(bne, RSA_F4)) {
+            break;
+        }
+
+        // generate rsa key
+        int res;
```
syan10, line 360: seems it's another common issues in your PR.
"res" is generally for "resource"
"ret" is for the return code, but seems not suitable for your case.

Change it to a valid name!

core/Enclave/enclave_hsm.cpp
```diff
@@ -294,43 +327,92 @@ sgx_status_t enclave_create_rsa_key(uint8_t *cmk_blob,
...
-        
-        BIO *public_key = BIO_new(BIO_s_mem());
-        BIO *private_key = BIO_new(BIO_s_mem());
+        if(!res) {
+            break;
+        }
 
-        PEM_write_bio_RSAPrivateKey(private_key, rsa, EVP_des_ede3_ofb(), (unsigned char*)g_domain_key, sizeof(g_domain_key), NULL, NULL);
+        // storage rsa key with pem format
+        public_key = BIO_new(BIO_s_mem());
```
syan10, line 379: why still malloc it in the enclave?

as i mentioned that you need to do it by the following steps:
1. calculate the memory size for the rsa key size
2. malloc it in the provider layer, and then pass the memory buffer into the enclave
3. enclave just check the memory is valid, and use it to store the new generated rsa keys.

core/Enclave/enclave_hsm.cpp
```diff
@@ -458,27 +540,111 @@ sgx_status_t enclave_rsa_encrypt(const uint8_t *cmk_blob,
...
+    if (keyspec == NULL || padding_mode == NULL)
+        return SGX_ERROR_INVALID_PARAMETER;
 
-    if(rsa == NULL)
+    uint32_t sealed_rsa_len;
+    switch(keyspec)
     {
-        printf("failed to load rsa key\n");
+        case EH_RSA_2048:
+            sealed_rsa_len = sgx_calc_gcm_data_size(0, 2105);
```
syan10, line 552: same here, hardcodes!

core/Enclave/enclave_hsm.cpp
```diff
@@ -458,27 +540,111 @@ sgx_status_t enclave_rsa_encrypt(const uint8_t *cmk_blob,
...
+
+            if (plaintext_len > RSA_OAEP_2048_SHA_256_MAX_ENCRYPTION_SIZE) {
+                printf("ecall rsa_encrypt plain len is up to 190B.\n");
```
syan10, line 555: seems that some place you use the function to get the max size, some place use the MACROs

core/Enclave/enclave_hsm.cpp
```diff
@@ -458,27 +540,111 @@ sgx_status_t enclave_rsa_encrypt(const uint8_t *cmk_blob,
...
+                return SGX_ERROR_INVALID_PARAMETER;
+            }
+
+            if (ciphertext == NULL || ciphertext_len < RSA_OAEP_2048_CIPHER_LENGTH) {
+                printf("ecall rsa_encrypt ciphertext len is too small.\n");
+                return SGX_ERROR_INVALID_PARAMETER;
+            }
+            break;
+        case EH_RSA_3072:
+            sealed_rsa_len = sgx_calc_gcm_data_size(0, 3060);
```
syan10, line 565: case for 4096?

core/Enclave/enclave_hsm.cpp
```diff
@@ -458,27 +540,111 @@ sgx_status_t enclave_rsa_encrypt(const uint8_t *cmk_blob,
...
+    BIO *bio = NULL;
+    RSA* rsa = NULL;
+    uint32_t public_key_size;
+    
+    do {
+        // load rsa public key
+        switch(keyspec)
+        {
+            case EH_RSA_2048:
+                public_key_size = 426;
```
syan10, line 606: same here

core/Enclave/enclave_hsm.cpp
```diff
@@ -458,27 +540,111 @@ sgx_status_t enclave_rsa_encrypt(const uint8_t *cmk_blob,
...
+                break;
+            case EH_RSA_3072:
+                public_key_size = 601;
+                break;
+        }
+        enc_key = (char*)malloc(cmk_blob_size);
+        sgx_gcm_decrypt(&g_domain_key,
```
syan10, line 613: sgx_gcm_decrypt is a wrapper for the sdk API implement by us, right?
if so, please rename it, don't use "sgx_xxx"

core/Enclave/enclave_hsm.cpp
```diff
@@ -489,30 +655,109 @@ sgx_status_t enclave_rsa_decrypt(const uint8_t *cmk_blob, size_t cmk_blob_size,
...
-    rsa_key_data_t *temp = (rsa_key_data_t*)cmk_blob;
-
-    BIO *bio = BIO_new_mem_buf(temp->rsa_private_key, -1);
-    RSA *rsa = PEM_read_bio_RSAPrivateKey(bio, NULL, NULL, g_domain_key);
-    
+    if (keyspec == NULL || padding_mode == NULL)
+        return SGX_ERROR_INVALID_PARAMETER;
 
-    if(rsa == NULL)
+    uint32_t sealed_rsa_len;
```
syan10, line 661: what's the meaning for "sealed_rsa_len"?

include/datatypes.h
```diff
@@ -139,16 +141,6 @@ typedef struct {
...
-typedef struct
-{
-    uint32_t    ciphertext_size;
-    uint32_t    aad_size;
-    uint8_t     iv[SGX_AESGCM_IV_SIZE];
-    uint8_t     mac[SGX_AESGCM_MAC_SIZE];
-    uint8_t     payload[];   /* ciphertext + aad */
-} aes_gcm_key_data_t;
-
 //rsa
```
syan10, line 144: you didn't change the definition of the structure? 

I think the first important thing is to define the structure, coding is just the pipeline work! if the structure not ready you may update this PR again and again...

core/App/ehsm_provider.cpp
whh39: They are the space that must be reserved for the padding mode, the rsa encryption plaintext length must be keysize-padding size.
I also look if there are the macros defined by sgx-ssl

core/App/ehsm_provider.cpp
whh39: There are some conflicts here. I hope to find the macro or function that has been defined. The three methods are reserved for the time being

core/App/ehsm_provider.cpp
whh39: 4096bits is going to be added soon

core/Enclave/enclave_hsm.cpp
```diff
@@ -109,19 +111,27 @@ void printf(const char *fmt, ...)
 
 static uint32_t sgx_calc_gcm_data_size(const uint32_t aad_size, const uint32_t plaintext_size)
 {
-    if (aad_size > UINT32_MAX - sizeof(aes_gcm_key_data_t))
+    if (aad_size > UINT32_MAX - sizeof(sgx_aes_gcm_data_ex_t))
```
yang8621, line 114: Where is the definition of sgx_aes_gcm_data_ex_t ?

core/Enclave/enclave_hsm.cpp
```diff
@@ -294,43 +327,92 @@ sgx_status_t enclave_create_rsa_key(uint8_t *cmk_blob,
...
+        BIO_read(public_key, p_public_key, BIO_pending(public_key));
+        BIO_read(private_key, p_private_key, BIO_pending(private_key));
 
-    return SGX_SUCCESS;
+        if (p_public_key == NULL && p_private_key == NULL)
+            break;
+
+        payload = string(p_public_key) + string(p_private_key);
+
+        sgx_gcm_encrypt(&g_domain_key, key_size, (uint8_t*)(payload.c_str()), 0, NULL, cmk_blob_size, (sgx_aes_gcm_data_ex_t *)cmk_blob);
```
yang8621, line 399: check return value

core/Enclave/enclave_hsm.cpp
```diff
@@ -294,43 +327,92 @@ sgx_status_t enclave_create_rsa_key(uint8_t *cmk_blob,
...
+    BIO_free(public_key);
+    BIO_free(private_key);
+    BN_free(bne);
+
+    SAFE_FREE(p_public_key);
+    SAFE_FREE(p_private_key);
+
+    string().swap(payload);
+    
+    ret = SGX_SUCCESS;
```
yang8621, line 413: always success?

core/Enclave/enclave_hsm.cpp
```diff
@@ -294,43 +327,92 @@ sgx_status_t enclave_create_rsa_key(uint8_t *cmk_blob,
...
+    if (keyspec == NULL || padding_mode == NULL)
+        return SGX_ERROR_INVALID_PARAMETER;
+
+    RSA     *rsa            = NULL;
+    BIGNUM  *bne            = NULL;
+    BIO     *public_key     = NULL;
+    BIO     *private_key    = NULL;
+    char    *p_public_key   = NULL;
+    char    *p_private_key  = NULL;
+    string payload;
```
yang8621, line 339: Can use memory directly?  string does not make sense.

core/Enclave/enclave_hsm.cpp
yang8621: check return value

core/Enclave/enclave_hsm.cpp
```diff
@@ -458,27 +540,111 @@ sgx_status_t enclave_rsa_encrypt(const uint8_t *cmk_blob,
...
-            RSA_public_encrypt(plaintext_len, plaintext, ciphertext, rsa, RSA_PKCS1_OAEP_PADDING);
+    if (plaintext == NULL || plaintext_len == 0) {
+        printf("ecall rsa_encrypt plaintext or len is wrong.\n");
+        return SGX_ERROR_INVALID_PARAMETER;
     }
 
-    return SGX_SUCCESS;
+    // encrypt
+    char* enc_key = NULL;
+    char* rsa_public_key = NULL;
```
yang8621, line 596: suggest using uint8_t*  but not char*

core/Enclave/enclave_hsm.cpp
```diff
@@ -489,30 +655,109 @@ sgx_status_t enclave_rsa_decrypt(const uint8_t *cmk_blob, size_t cmk_blob_size,
...
+            case EH_RSA_2048:
+                private_key_size = 1679;
+                public_key_size = 426;
+                break;
+            case EH_RSA_3072:
+                private_key_size = 2459;
+                public_key_size = 601;
+                break;
+        }
+        sgx_gcm_decrypt(&g_domain_key,
```
yang8621, line 713: check

core/Enclave/enclave_hsm.cpp
```diff
@@ -458,27 +540,111 @@ sgx_status_t enclave_rsa_encrypt(const uint8_t *cmk_blob,
...
+        rsa = PEM_read_bio_RSAPublicKey(bio, NULL, NULL, NULL);
+        if (rsa == NULL) {
+            printf("failed to load rsa key\n");
+            break;
+        }
+
+        // make encryption
+        int res = RSA_public_encrypt(plaintext_len, plaintext, ciphertext, rsa, padding_mode);
+
+        if (!res)
```
yang8621, line 635: RSA_public_encrypt() returns the size of the encrypted data (i.e., RSA_size(rsa)).
On error, -1 is returned;

core/Enclave/enclave_hsm.cpp
```diff
@@ -458,27 +540,111 @@ sgx_status_t enclave_rsa_encrypt(const uint8_t *cmk_blob,
...
+            if (plaintext_len > RSA_OAEP_3072_SHA_256_MAX_ENCRYPTION_SIZE) {
+                printf("ecall rsa_encrypt plain len is up to 318B.\n");
+                return SGX_ERROR_INVALID_PARAMETER;
+            }
+
+            if (ciphertext == NULL || ciphertext_len < RSA_OAEP_3072_CIPHER_LENGTH) {
+                printf("ecall rsa_encrypt ciphertext len is too small.\n");
+                return SGX_ERROR_INVALID_PARAMETER;
+            }
+            break;
```
yang8621, line 576: no default?

core/Enclave/enclave_hsm.cpp
```diff
@@ -489,30 +655,109 @@ sgx_status_t enclave_rsa_decrypt(const uint8_t *cmk_blob, size_t cmk_blob_size,
...
+        }
+        
+        if (plaintext_len == 0) {
+            int plaintext_len;
+            // compute plaintext length
+            unsigned char* temp_plaintext = (unsigned char*)malloc(RSA_size(rsa)); // malloc the enough space for RSA_private_decrypt
+            
+            plaintext_len = RSA_private_decrypt(ciphertext_len, ciphertext, temp_plaintext, rsa, padding_mode); // will return the plaintext length after decrypt 
+            
+            *req_plaintext_len = plaintext_len - 1; // delete '\0' for char type
```
yang8621, line 739: I do not believe this is required.   The plaintext should be always regarded as binary/uint8_t[n] but not a string/char*.

core/Enclave/enclave_hsm.cpp
```diff
@@ -489,30 +655,109 @@ sgx_status_t enclave_rsa_decrypt(const uint8_t *cmk_blob, size_t cmk_blob_size,
...
+
+            SAFE_FREE(temp_plaintext);
+
+            ret = SGX_SUCCESS;
+        } else {    
+            // make decryption
+            int res = RSA_private_decrypt(ciphertext_len, ciphertext, plaintext, rsa, padding_mode);       
+            
+            if (!res)
```
yang8621, line 748: same

core/App/ehsm_provider.cpp
whh39: yes

core/Enclave/enclave_hsm.cpp
whh39: yes
I'll add some soon

core/Enclave/enclave_hsm.cpp
whh39: copy understood

core/Enclave/enclave_hsm.cpp
whh39: yes, I'll change it

core/Enclave/enclave_hsm.cpp
whh39: Here is just use some memory to storage public key and private key and use as the buffer for keyblob with aes_gcn_128 encryption 
the space will be free in enclave

core/Enclave/enclave_hsm.cpp
whh39: > check return value

yes, I'll add

> sgx_gcm_decrypt is a wrapper for the sdk API implement by us, right? if so, please rename it, don't use "sgx_xxx"

the function name will change.

core/Enclave/enclave_hsm.cpp
whh39: it will be changed

include/datatypes.h
whh39: Here, for rsa, it will use the same struct as aes_gcm encryption.
The rsa keypair will be storage into payload[]

core/Enclave/enclave_hsm.cpp
whh39: in datatypes.h, the same struct as before

core/Enclave/enclave_hsm.cpp
whh39: copy understood

core/Enclave/enclave_hsm.cpp
whh39: I have find this problem, will fix soon

core/Enclave/enclave_hsm.cpp
whh39: string here is used for a strcat for rsa keypair, I can modify it if necessary

core/Enclave/enclave_hsm.cpp
whh39: copy

core/Enclave/enclave_hsm.cpp
whh39: Reruen check here does have a problem, I'll check it

core/Enclave/enclave_hsm.cpp
whh39: Yes, they are same, I'll change it

core/App/ehsm_napi.cpp
```diff
@@ -247,6 +256,7 @@ char* NAPI_Encrypt(const char* paramJson)
     int plaintext_len = plaintext_str.size();
     int aad_len = aad_str.size();
 
+    printf("cmk_len=%d\n", cmk_len);
```
yang8621, line 259: remove debug info, please

include/datatypes.h
```diff
@@ -221,4 +212,46 @@ typedef enum {
...
+        case RSA_PKCS1_OAEP_PADDING:
+            return RSA_PKCS1_OAEP_PADDING_SIZE;
+        case RSA_PKCS1_PADDING:
+            return RSA_PKCS1_PADDING_SIZE;
+        default:
+            return 0;
+    }
+}
+
+static uint32_t ehsm_get_rsa_max_encryption_size(const uint32_t keyspec, const uint32_t padding_mode)
```
yang8621, line 227: It's used in other files, why define it as static?
And same @line 241

include/datatypes.h
```diff
@@ -35,6 +35,9 @@
 #include "sgx_dh.h"
 #include "sgx_tseal.h"
 
+#include "openssl/rsa.h"
+#include "openssl/evp.h"
```
yang8621, line 39: Can dkeycache and dkeyserver make pass?

core/Enclave/enclave_hsm.cpp
```diff
@@ -107,21 +94,59 @@ void printf(const char *fmt, ...)
...
+    switch (keyspec)
+    {
+        case EH_RSA_2048:
+            return RSA_2048_PUBLIC_KEY_PEM_SIZE + RSA_2048_PRIVATE_KEY_PEM_SIZE;
+        case EH_RSA_3072:
+            return RSA_3072_PUBLIC_KEY_PEM_SIZE + RSA_3072_PRIVATE_KEY_PEM_SIZE;
+        case EH_RSA_4096:
+            return RSA_4096_PUBLIC_KEY_PEM_SIZE + RSA_4096_PRIVATE_KEY_PEM_SIZE;
+        default:
+            return UINT32_MAX;
```
yang8621, line 133: if the default error is UINT32_MAX, every places should check the return value.
@322, @704

yang8621: Is 0 a valid value?  If not, you can use 0 as the default value to indicate an error.

core/Enclave/enclave_hsm.cpp
```diff
@@ -458,27 +554,103 @@ sgx_status_t enclave_rsa_encrypt(const uint8_t *cmk_blob,
...
+    BIO *bio = NULL;
+    RSA* rsa_pubkey = NULL;
+    uint32_t public_key_size;
+    
+    do {
+        // load rsa public key
+        //
+        public_key_size = ehsm_get_rsa_public_key_pem_size(keyspec);
+        
+        enc_key = (uint8_t*)malloc(cmk_blob_size);
```
yang8621, line 615: enc_key is the plaintext to be decrypted.
cmk_blob_size is the size of cmk_blob. The two have nothing relationship.

yang8621: you can refer to https://github.com/intel/ehsm/blob/c665aa19f89988080493a6c82913d7b2af73e9d8/core/Enclave/enclave_hsm.cpp#L142
For GCM, the ciphertext len is same to plaintext len. you can use it to get the plaintext len.

core/Enclave/enclave_hsm.cpp
```diff
@@ -225,12 +248,25 @@ static sgx_status_t sgx_gcm_encrypt(const sgx_aes_gcm_128bit_key_t *key,
  * sgx_rijndael128GCM_decrypt() is running in enclave
  * 
  */
-static sgx_status_t sgx_gcm_decrypt(/* param */)
+static sgx_status_t ehsm_gcm_decrypt(const sgx_aes_gcm_128bit_key_t *key,
+                                    uint32_t *plaintext_size, uint8_t *plaintext,
```
yang8621, line 252: [Why is plaintext_size defined as a pointer? Unless the caller does not know the size of plaintext, and have to call the function at the 1st time to get the size.  Otherwise, no need to define it as a pointer.

core/Enclave/enclave_hsm.cpp
```diff
@@ -458,27 +554,103 @@ sgx_status_t enclave_rsa_encrypt(const uint8_t *cmk_blob,
...
+        public_key_size = ehsm_get_rsa_public_key_pem_size(keyspec);
+        
+        enc_key = (uint8_t*)malloc(cmk_blob_size);
+        ret = ehsm_gcm_decrypt(&g_domain_key,
+                              (uint32_t*)&cmk_blob_size, enc_key,
+                              (sgx_aes_gcm_data_ex_t *)cmk_blob);
+        if (ret != SGX_SUCCESS)
+            break;
+
+        rsa_public_key = (uint8_t*)malloc(public_key_size);
```
yang8621, line 622: why the decryption of cmk_blob is rsa public key?  
Can you show the structure of cmk_blob you defined, like https://github.com/intel/ehsm/blob/main/core/Enclave/enclave_hsm.cpp#L479
I think this is the fundamental step we have to align.

core/Enclave/enclave_hsm.cpp
```diff
@@ -458,27 +554,103 @@ sgx_status_t enclave_rsa_encrypt(const uint8_t *cmk_blob,
                                  uint8_t keyspec,
                                  uint8_t padding_mode)
```
yang8621, line 555: are keyspec and pading_mode still required for encryption/decryption?
I assume the information should be recorded in cmk_blob.
@syan10

core/Enclave/enclave_hsm.cpp
```diff
@@ -281,56 +317,116 @@ sgx_status_t enclave_create_rsa_key(uint8_t *cmk_blob,
...
+        BIO_read(private_key, (char*)p_private_key, private_key_len);
 
-    return SGX_SUCCESS;
+        if (p_public_key == NULL && p_private_key == NULL)
+            break;
+
+        payload = (uint8_t*)malloc(public_key_len + private_key_len + 1);
+        memcpy(payload, p_public_key, public_key_len);
+        memcpy(payload + public_key_len, p_private_key, private_key_len);
+
```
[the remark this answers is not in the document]
yang8621: And I dont think public key must be encrypted.

core/Enclave/enclave_hsm.cpp
yang8621: OK   understood
The head bytes are public key, and the tail bytes are private key.

core/Enclave/enclave_hsm.cpp
```diff
@@ -458,27 +554,103 @@ sgx_status_t enclave_rsa_encrypt(const uint8_t *cmk_blob,
...
+        if (rsa_pubkey == NULL) {
+            printf("failed to load rsa key\n");
+            break;
+        }
+
+        // make encryption
+        //
+        int encryption_len = RSA_public_encrypt(plaintext_len, plaintext, ciphertext, rsa_pubkey, padding_mode);
+
+        if (encryption_len == -1 || encryption_len != ciphertext_len)
```
yang8621, line 641: see @583, ciphertext_len can be big enough

yang8621:
Suggest referring to the code:
https://github.com/intel/ehsm/blob/main/core/Enclave/enclave_hsm.cpp#L478 ~ L795
and define a reasonable rsa_cmk_blob { }

core/Enclave/enclave_hsm.cpp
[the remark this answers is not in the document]
yang8621: BTW, enclave codes should use memcpy_s

core/App/ehsm_napi.cpp
whh39: sorry, I'll

include/datatypes.h
whh39: statis function to avoid multiple definition

include/datatypes.h
whh39: sorry it's a mistake, I have fixed

core/Enclave/enclave_hsm.cpp
whh39: This function will use in ehsm_calc_gcm_data_size(add_size, plaintext_size) to get keybloblen
And UINT32_MAX in it will return error

core/Enclave/enclave_hsm.cpp
whh39: copy, i will check this

core/Enclave/enclave_hsm.cpp
whh39: Except for the function name, I haven't made any changes. The previous definition is like this. I also tend to modify it

core/Enclave/enclave_hsm.cpp
whh39: yes， and BIO_read will use the rsa key head to get keypair

core/Enclave/enclave_hsm.cpp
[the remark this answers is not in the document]
whh39: > BTW, enclave codes should use memcpy_s

copy understood

core/Enclave/enclave_hsm.cpp
whh39: After calling RSA_public_encrypt() correctly, the return value should be RSA_ Size(), which is written in OpenSSL code to check whether it is called correctly.
I'll check this part again

core/App/ehsm_napi.cpp
```diff
@@ -1135,10 +1145,9 @@ char* NAPI_AsymmetricDecrypt(const char* paramJson)
     switch (cmk.metadata.keyspec)
     {
         case EH_RSA_2048:
-            // TODO
-            break;
         case EH_RSA_3072:
-            ciphertext_maxLen = RSA_OAEP_3072_CIPHER_LENGTH;
+        case EH_RSA_4096:
+            ciphertext_maxLen = ehsm_get_rsa_cipher_len(cmk.metadata.keyspec);
```
syan10, line 1150: seems internal interfaces not suitable to call ehsm_xxx

core/App/ehsm_provider.h
```diff
@@ -44,28 +44,12 @@
...
-#define EH_AES_GCM_MAC_SIZE 16
-
-#define RSA_OAEP_2048_SHA_256_MAX_ENCRYPTION_SIZE       190
-// #define RSA_2048_OAEP_SHA_1_MAX_ENCRYPTION_SIZE       214
-
-#define RSA_OAEP_3072_SHA_256_MAX_ENCRYPTION_SIZE       318
-// #define RSA_3072_OAEP_SHA_1_MAX_ENCRYPTION_SIZE       342
-
-#define SM2PKE_MAX_ENCRYPTION_SIZE                      6047
+#define EH_CMK_MAX_SIZE (8*1024)
```
syan10, line 51: why define so large size?

core/Enclave/enclave_hsm.cpp
```diff
@@ -49,34 +50,20 @@
...
-#define RSA_OAEP_2048_SHA_256_MAX_ENCRYPTION_SIZE       190
-//#define RSA_2048_OAEP_SHA_1_MAX_ENCRYPTION_SIZE       214
+#define RSA_2048_KEY_BITS   2048
+#define RSA_3072_KEY_BITS   3072
+#define RSA_4096_KEY_BITS   4096
 
-#define RSA_OAEP_3072_SHA_256_MAX_ENCRYPTION_SIZE       318
-//#define RSA_3072_OAEP_SHA_1_MAX_ENCRYPTION_SIZE       342
+#define RSA_2048_PUBLIC_KEY_PEM_SIZE    426
+#define RSA_2048_PRIVATE_KEY_PEM_SIZE    1679
```
syan10, line 60: how did you get the size? any link?

core/Enclave/enclave_hsm.cpp
```diff
@@ -107,21 +94,59 @@ void printf(const char *fmt, ...)
     ocall_print_string(buf);
 }
 
-static uint32_t sgx_calc_gcm_data_size(const uint32_t aad_size, const uint32_t plaintext_size)
+static uint32_t ehsm_calc_gcm_data_size(const uint32_t aad_size, const uint32_t plaintext_size)
```
syan10, line 97: same here, the name issue

core/Enclave/enclave_hsm.cpp
syan10: the structure is defined by ourselves? so not name it as sgx_xxx

core/Enclave/enclave_hsm.cpp
```diff
@@ -107,21 +94,59 @@ void printf(const char *fmt, ...)
...
         return UINT32_MAX;
 
-    if (sizeof(aes_gcm_key_data_t) > UINT32_MAX - plaintext_size - aad_size)
+    if (sizeof(sgx_aes_gcm_data_ex_t) > UINT32_MAX - plaintext_size - aad_size)
+        return UINT32_MAX;
+
+    return (aad_size + plaintext_size + sizeof(sgx_aes_gcm_data_ex_t));
+}
+
+static uint32_t sgx_get_gcm_ciphertext_size(const sgx_aes_gcm_data_ex_t *gcm_data)
```
syan10, line 114: same here

core/Enclave/enclave_hsm.cpp
```diff
@@ -107,21 +94,59 @@ void printf(const char *fmt, ...)
...
-    return (aad_size + plaintext_size + sizeof(aes_gcm_key_data_t));
+    return gcm_data->ciphertext_size;
+}
+
+static uint32_t ehsm_get_rsa_key_pem_size(const uint32_t keyspec)
+{
+    switch (keyspec)
+    {
+        case EH_RSA_2048:
+            return RSA_2048_PUBLIC_KEY_PEM_SIZE + RSA_2048_PRIVATE_KEY_PEM_SIZE;
```
syan10, line 127: so the pem size is the whole size of the keyblob which encrypted by the domain key?

whh39: It is pem size + the size of the struct for aes_gcm_128 encryption

struct 
{
    uint32_t  ciphertext_size;
    uint32_t  aad_size;
    uint8_t   reserve1[8];
    uint8_t   iv[SGX_AESGCM_IV_SIZE];
    uint8_t   reserve2[4];
    uint8_t   mac[SGX_AESGCM_MAC_SIZE];
    uint8_t   payload[];   /* rsa keypair pem */
}

core/Enclave/enclave_hsm.cpp
yang8621: The previous usage needs to call the function twice.  The 1st time calling is to get the size of the plaintext

core/Enclave/enclave_hsm.cpp
```diff
@@ -281,56 +317,116 @@ sgx_status_t enclave_create_rsa_key(uint8_t *cmk_blob,
...
+
+    // check cmk_blob and cmk_blob_size
+    //
     if (cmk_blob == NULL || cmk_blob_size < real_keyblob_size) {
         printf("ecall create_rsa_key cmk_keyblob_size:%lu < key_blob_size:%d.\n", cmk_blob_size, real_keyblob_size);
         return SGX_ERROR_INVALID_PARAMETER;
     }
-        rsa_key_data_t *keyblob = (rsa_key_data_t *)cmk_blob;
-     
+    if (keyspec == NULL || padding_mode == NULL)
```
yang8621, line 341: Where is padding_mode used in create rsa key?

core/App/ehsm_provider.h
whh39: previous value 4*1024 is not enough for rsa4096

PR #195, opened 2022-11-29
Change domain key from 128bits to 256bits

whh39:
Change domain key from 128bits to 256bits


dkeyserver/Enclave/enclave.cpp
```diff
@@ -49,7 +49,7 @@
 #include "byteswap.h"
 #include "openssl_utility.h"
 
-#define SGX_DOMAIN_KEY_SIZE 16
+#define SGX_DOMAIN_KEY_SIZE 32
```
syan10, line 52: 删掉重复的定义， 应该都可以include "datatypes.h"

PR #196, opened 2022-11-29
Add safe-test for asymmetric key pair

whh39:
Add safe-test for asymmetric key pair

the safe-test will support for:
rsa: generate key pair
ecc: generate key pair
sm2: generate key pair

### DIFF
--- a/core/Enclave/key_factory.cpp
+++ b/core/Enclave/key_factory.cpp
@@ -113,7 +113,7 @@ sgx_status_t ehsm_create_keyblob(uint8_t *plaintext,
                           NULL, 0,
                           keyblob_data->iv, SGX_AESGCM_IV_SIZE,
                           keyblob_data->mac, SGX_AESGCM_MAC_SIZE);
-                          
+
     if (SGX_SUCCESS != ret)
         printf("gcm encrypting failed.\n");
     else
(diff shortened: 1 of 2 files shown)


core/Enclave/key_factory.cpp
```diff
@@ -261,6 +291,13 @@ sgx_status_t ehsm_create_rsa_key(ehsm_keyblob_t *cmk)
     if (rsa_keypair == NULL)
         goto out;
 
+#ifdef FIPS_MODE
```
yang8621, line 294: is FIPS_MODE defined?

core/Enclave/key_factory.cpp
```diff
@@ -217,6 +217,36 @@ sgx_status_t ehsm_create_aes_key(ehsm_keyblob_t *cmk)
     return ret;
 }
 
+#ifdef FIPS_MODE
+static bool test_RSA_keypair(RSA *keypair)
+{
```
syan10, line 222: test_RSA_keypair -> pair_wise_test_for_rsa

FIPS_MODE -> ENABLE_PAIR_WISE_TEST

core/Enclave/key_factory.cpp
whh39: in makefile

core/Enclave/key_factory.cpp
whh39: ok

core/Enclave/key_factory.cpp
whh39: it's a mistake, FIPS_MODE will remove

PR #199, opened 2022-12-05
Fix the memory leak issue

whh39:
Fix the memory leak issue

free the memory after use:
1. sm2_crypto_test()
2. ehsm_rsa_decrypt()


core/Enclave/key_operation.cpp
```diff
@@ -738,7 +738,7 @@ sgx_status_t ehsm_rsa_decrypt(const ehsm_keyblob_t *cmk,
 
     if (plaintext->datalen == 0)
```
syan10, line 739: commit msg
"
Fix the memory leak issue

free the memory after use:
1. sm2_crypto_test()
2. ehsm_rsa_decrypt()

"

syan10 (approved):

PR #201, opened 2022-12-05
Use calloc to zero the memory of apikey buffer

Yidingluo1:
test: passed the unit test

Signed-off-by: luoyiding <yidingx.luo@intel.com>

### DIFF
--- a/core/App/ffi_operation.cpp
+++ b/core/App/ffi_operation.cpp
@@ -1432,7 +1432,7 @@ extern "C"
         }
         appid->datalen = UUID_STR_LEN;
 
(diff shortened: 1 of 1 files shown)


core/App/ffi_operation.cpp
```diff
@@ -1432,7 +1432,7 @@ extern "C"
         }
         appid->datalen = UUID_STR_LEN;
```
syan10, line 1433: commit msg
"
Use calloc to zero the memory of apikey buffer

test: passed the unit test
Signed-off-by: luoyiding <yidingx.luo@intel.com>
"

Yidingluo1: done

PR #211, opened 2022-12-13
Return the sgx error code directly for verifyQuote

Yidingluo1:
test: passed the unit test

Signed-off-by: yidingluo <yidingx.luo@intel.com>


core/App/ehsm_provider.cpp
```diff
@@ -791,7 +791,7 @@ ehsm_status_t GenerateQuote(ehsm_data_t *quote)
 ehsm_status_t VerifyQuote(ehsm_data_t *quote,
                           const char *mr_signer,
                           const char *mr_enclave,
```
syan10, line 793: title:
Return the sgx error code directly for verifyQuote

Yidingluo1: done

syan10 (approved):

PR #220, opened 2022-12-16
Mount SGX device for K8S deployment

wanghouqi:
Signed-off-by: wanghouqi <houqix.wang@intel.com>

### DIFF
--- a/docker/k8s/ehsm-kms.yaml
+++ b/docker/k8s/ehsm-kms.yaml
@@ -156,6 +156,15 @@ spec:
         name: dkeycache
     spec:
       volumes:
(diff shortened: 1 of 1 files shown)


docker/k8s/ehsm-kms.yaml
```diff
@@ -156,6 +156,15 @@ spec:
         name: dkeycache
```
syan10, line 156: "Mount SGX device for K8S deployment"

wanghouqi: fix

syan10 (approved):

PR #226, opened 2022-12-21
Use the log_x after the logger initialized

hongjingmao:
Signed-off-by: hongjingmao <hongjing.mao@intel.com>


core/App/ehsm_provider.cpp
```diff
@@ -51,6 +51,7 @@
 #include "sgx_ql_quote.h"
 #include "sgx_dcap_quoteverify.h"
```
syan10, line 52: commit msg
"
Use the log_x after the logger initialized
"

syan10 (approved):

PR #233, opened 2022-12-27
KMS supports run mode

wanghouqi:
- KMS supports two startup modes, single and cluster, The cluster is
default mode.
      single mode: You can run KMS directly without DKeyCache, KMS will be use a default DK.
      cluster mode: You must connect to DKeyCache run KMS. If you cannot connect DKeyCache, KMS cannot start.
 - KMS server add config.yml to set server related configuration
information.
 - You can set the server parameters in three ways: nodejs argv, system env and
config.yml, Their priority is nodejs argv > system env > config.yml

test: passed the k8s cluster

Signed-off-by: wanghouqi <houqix.wang@intel.com>


core/App/ehsm_provider.cpp
```diff
@@ -61,6 +64,8 @@
 
 using namespace std;
 
+#define FILE_NAME (std::string(RUNTIME_FOLDER) + "single_test_dkey.bin").c_str()
```
syan10, line 67: FILE_NAME -> DKEY_FILE_NAME

wanghouqi: fixed

PR #245, opened 2023-03-03
Add running mode support

syan10:
KMS supports two running modes: single and cluster
- single mode: You can run KMS directly without DKeyCache, KMS will be create a DK by self
- cluster mode: You must connect to DKeyCache run KMS. If you cannot connect DKeyCache, KMS cannot start. The cluster is default mode.

you can change the default running mode by setting the $run_mode through: nodejs argv, system env or config.yml.
Their priority is nodejs argv > system env > config.yml(prefered)

test: passed the k8s cluster


ehsm_kms_service/config.yml
```diff
@@ -0,0 +1,35 @@
+# KMS Server base config.
+service:
+  # KMS service's port, you can set an integer, the default value is 9000.
+  port: 9001
+  # KMS service running mode, you can set single or cluster, the default value is cluster.
+  #    single mode: You can run KMS directly without DKeyCache, KMS will be create a DK by self.
+  #    cluster mode: You must connect to DKeyCache run KMS. If you cannot connect DKeyCache, KMS cannot start.
+  run_mode: "single"
```
yang8621, line 8: cluster?

syan10: updated

yang8621 (approved):

PR #279, opened 2023-07-19
add initial support for go client sdk

whh39:
currently only support createkey, getpublickey, sign, verify, more interfaces will be added later.

### DIFF
--- a/sdk/go/client.go
+++ b/sdk/go/client.go
@@ -1,0 +1,93 @@
+package ehsm
+
+import (
+    tls "crypto/tls"
+    "encoding/json"
+    "fmt"
+    http "net/http"
+    "sort"
+    "os"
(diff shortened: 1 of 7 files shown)


syan10 (approved):

sdk/go/client.go
```diff
@@ -0,0 +1,91 @@
...
+
+type Client struct {
+    modifyLock              sync.RWMutex
+    addr                    string
+    apikey                  string
+    appid                   string
+    action                  string
+    resquest                *orderedmap.OrderedMap
+}
+
```
yang8621, line 24: For production environment, the web server should request a formally issued
certificate, and change the below skipVerify to False.

whh39: ok

PR #288, opened 2023-07-28
update nonce in ehsm_kms_service

ForLisiteng:
no longer use the timestamp as nonce, since it will have concurrency access issues, even for different requests.

now, the nonce is optional which max length is 64 bytes, the caller can decide whether to pass it with the request data.

Signed-off-by: lisiteng <sitengx.li@intel.com>


ehsm_kms_service/function.js
```diff
@@ -431,7 +431,7 @@ const _checkParams = function (req, res, next, nonce_database, DB) {
         }
         const {
             appid,
-            timestamp: nonce,
```
syan10, line 434: add a comment to describe nonce is a optional param

syan10 (approved):

yang8621:
is undefined checked? 
there are both null and undefined for js.

PR #294, opened 2023-08-18
bug fix for AsymmetricDecrypt/AsymmetricDecrypt in kms

whh39:


ehsm_kms_service/router.js
```diff
@@ -206,17 +206,19 @@ const router = async (p) => {
       break
     case KMS_ACTION.cryptographic.AsymmetricEncrypt:
       try {
-        const { keyid, plaintext } = payload
+        let { keyid, plaintext, padding_mode } = payload
+        padding_mode = ehsm_padding_mode_t[padding_mode]
         const cmk_base64 = await find_cmk_by_keyid(appid, keyid, res, DB)
```
syan10, line 211: this padding_mode not required right? if user not provide it what will be happen?

yang8621: 是的   这样写也会越界

syan10 (approved):

PR #296, opened 2023-08-22
Feat: move HMAC verification into enclave

### LINKED ISSUE #274: move the HMAC verification into core enclave
currently, the request params' hmac verification process is did in web service layer, which may lead to the apikey exposure, need to move it into core enclave to enhance the security.

truc0:
1. add `gen_hmac_safe()` function in `functions.js` and use it for verifying signature of incoming request
2. add `EH_GEN_HMAC` action and `ffi_generateHmac()` handler
3. add `enclave_generate_hmac()` function to enclave part of `core` app for handling HMAC generation in enclave
4. the original `gen_hmac()` function is still kept since `verifyQuote()` depends on this function

close #274


core/App/ehsm_provider.cpp
```diff
@@ -293,6 +293,9 @@ uint32_t EHSM_FFI_CALL(const char *reqJson, char *respJson)
     case EH_VERIFY_QUOTE:
         ffi_verifyQuote(payloadJson, respJson);
         break;
```
syan10, line 295: please combine the 2 commits into one

truc0: Sure, the commits should be squashed. Please check, thanks.

core/App/ehsm_provider.h
```diff
@@ -324,4 +325,16 @@ apikey -- 32byte random number
 */
 ehsm_status_t Enroll(ehsm_data_t *appid, ehsm_data_t *apikey);
 
+/*
+Description:
+Generate HMAC(SHA-256) of payload with given apikey
+Input:
+cmk -- A symmetric cmk
+apikey -- 32byte random number (in plaintext, not encrypted)
```
syan10, line 333: it should be in ciphertext

core/App/ehsm_provider.h
```diff
@@ -324,4 +325,16 @@ apikey -- 32byte random number
...
+payload -- the payload, a string
```
syan10, line 334: payload is based64 encoded string of the request data

core/App/ffi_operation.cpp
```diff
@@ -1535,4 +1535,82 @@ extern "C"
...
+        // output params
+        ehsm_data_t *hmac = NULL;
+        std::string hmac_str;
+        RetJsonObj retJsonObj;
+
+        // 0. prepare data
+        JSON2STRUCT(payloadJson, cmk);
+        JSON2STRUCT(payloadJson, apikey);
+
+        payload_str = payloadJson.readData_string("payload");
```
syan10, line 1574: check if null?

core/App/ffi_operation.cpp
```diff
@@ -1535,4 +1535,82 @@ extern "C"
...
+        payload_size = payload_str.size();
+        payload = (ehsm_data_t *)malloc(APPEND_SIZE_TO_DATA_T(payload_size));
+        if (payload == NULL)
+        {
+            retJsonObj.setCode(retJsonObj.CODE_BAD_REQUEST);
+            retJsonObj.setMessage("The payload's length is invalid.");
+            goto out;
+        }
+        payload->datalen = payload_size;
+        memcpy_s(payload->data, payload_size, (uint8_t *)payload_str.data(), payload_size);
```
syan10, line 1584: payload_str not do base64 decode, because the user side also do the  hmac with b64 string, right?  you can add some comments here, because it's some  difference with the other functions

core/Enclave/enclave_hsm.cpp
```diff
@@ -940,4 +940,73 @@ sgx_status_t enclave_verify_quote_policy(uint8_t *quote, uint32_t quote_size,
...
+    ehsm_data_t *aad = NULL;
+    ehsm_data_t *rawApiKey = NULL;
+
+    // create empty aad
+    aad = (ehsm_data_t *)malloc(APPEND_SIZE_TO_DATA_T(0));
+    if (aad == NULL)
+    {
+        return SGX_ERROR_UNEXPECTED;
+    }
+    aad->datalen = 0;
```
syan10, line 988: even if you don't want to set aad, still need to malloc it?  seems the existing check logic of aad has some issue in current codes?

core/Enclave/enclave_hsm.cpp
```diff
@@ -940,4 +940,73 @@ sgx_status_t enclave_verify_quote_policy(uint8_t *quote, uint32_t quote_size,
...
+        log_w("apikey decrypt failed");
+        goto out;
+    }
+
+    // 2. Generate HMAC
+    ret = sgx_hmac_sha256_msg(payload->data, payload->datalen, rawApiKey->data, rawApiKey->datalen, hmac->data, hmac->datalen);
+
+out:
+    SAFE_FREE(aad);
+    SAFE_FREE(rawApiKey);
```
yang8621, line 1010: before free the sensitive data, better to clear it by memset_s(); 
you can refer to the samples in this same file.

core/App/ehsm_provider.h
truc0: Thanks for review! This is resolved in the latest commit.

core/App/ehsm_provider.h
truc0: Currently the `payload` sent by `ehsm_key_server/function.js` is a pure string (without base64 encode). 

The `gen_hmac()` function and related functions are only called when verifying incoming request and verifying quote in NodeJS side.  It accepts an object and turn it into string by `params_sort_str()` function. Thus, the original `gen_hmac()` does not encode the string into base64 format before calculating HMAC.

Th implementation of `gen_hmac_safe()` uses the same logic to ensure the output of `gen_hac()` and `gen_hmac_safe()` is consistant. Should I encode it in the NodeJS side and decode it in the ehsm_core?

core/App/ffi_operation.cpp
truc0: Thanks for reviewing! The validation of `cmk` and `apikey` should be added in the latest commit, please chceck.

core/Enclave/enclave_hsm.cpp
truc0: Thanks for reviewing! The sensitive info (`rawApiKey` in this function) should be set to 0 before returning, please check.

core/Enclave/enclave_hsm.cpp
truc0: Thanks for reviewing!

The `enclave_decrypt` function does allow passing `aad = NULL` with `aad_size = 0`. However, all tests related to `CreateKey` were failed when passing `aad = NULL` to the `enclave_decrypt` function. I am still investigating the problem.

truc0: I have traced the call chain of `enclave_decrypt` funciton and found the bug.

The `ehsm_aes_gcm_decrypt` treats `aad` as a non-null variable, directly uses `aad->data` without checking if `aad` is `NULL`.

### Investigation steps

The `enclave_generate_hmac` returns `SGX_ERROR_ENCLAVE_CRASH (0x1006)` when passing `aad=NULL` to `enclave_decrypt` in `enclave_generate_hmac` function.

This is because the `cmk` use `EH_AES_GCM_192` as keyspec, which transfer the control flow to `ehsm_aes_gcm_decrypt()`. However, the `ehsm_aes_gcm_decrypt()` does not check the `aad` varible before the following usage:

```cpp
// https://github.com/intel/ehsm/blob/main/core/Enclave/key_operation.cpp#L390
    ret = aes_gcm_decrypt(key,
                          plaintext->data,
                          block_mode,
                          cipherblob->data,
                          cipherblob->datalen - EH_AES_GCM_IV_SIZE - EH_AES_GCM_MAC_SIZE,
                          aad->data,  // BUG: DEREFERENCE OF NULL
                          aad->datalen, // BUG: DEREFERENCE OF NULL
                          iv,
                          SGX_AESGCM_IV_SIZE,
                          l_tag,
                          SGX_AESGCM_MAC_SIZE);
``` 

Then the enclave is crashed.

### Possible fix

A simple but work fix is changing the call to `aes_gcm_decrypt` in `ehsm_aes_gcm_decrypt` to the following:

```cpp
    ret = aes_gcm_decrypt(key,
                          plaintext->data,
                          block_mode,
                          cipherblob->data,
                          cipherblob->datalen - EH_AES_GCM_IV_SIZE - EH_AES_GCM_MAC_SIZE,
                          aad ? aad->data : NULL,
                          aad ? aad->datalen : NULL,
                          iv,
                          SGX_AESGCM_IV_SIZE,
                          l_tag,
                          SGX_AESGCM_MAC_SIZE);
```

Tests passed after the modification above.

Shall I open a new issue / PR for fixing this?

core/App/ffi_operation.cpp
truc0: Thanks for reviewing!

According to the original implementation of `gen_hmac()`, the user side does the hmac with pure string.

The string used for hmac is constructed by concatenating `sign_params` to format like `key1=value1&key2=value2` . The implementation is in [ehsm/ehsm_kms_server/function.js](https://github.com/intel/ehsm/blob/c9a8ea0248ad2457cd2d24fbb8c1c725566036ee/ehsm_kms_service/function.js#L357).

core/Enclave/enclave_hsm.cpp
yang8621: Thanks for the root cause. Of course, please create a separate PR to fix it.  Thanks.

core/App/ehsm_provider.h
syan10: yes, please do it in this way

core/Enclave/enclave_hsm.cpp
syan10: yes, agree

truc0: The new PR #302 is created.

truc0: The fix patch of #302 has been applied to current branch in the latest commit.

More detailly, `NULL` is passed as `aad` and `0` is passed as `aad_size` when calling `enclave_decrypt` function in `enclave_generate_hmac` function instead of allocating `sizeof(ehsm_data_t)` for `aad` without storing any data.

yang8621:
Integrate into one commit.

core/App/ehsm_provider.h
syan10: already fixed?

ehsm_kms_service/function.js
```diff
@@ -601,33 +601,52 @@ const _query_api_key = async (DB, appid) => {
...
 
 /**
- * Gen Hmac 
+ * Gen Hmac of sign_params safely, without transfering plaintext of apikey outside enclave
  * @param {object} DB //Database object
  * @param {string} appid // APP ID from client
  * @param {object} sign_params // A list object contain string params
  * @returns {object} (error, hmac) // One object contain "error" and "hmac" attributes
  */
 const gen_hmac = async (DB, appid, sign_params) => {
```
syan10, line 610: not updated? using gen_hmac_safe?

core/App/ehsm_provider.h
truc0: Yes. Now the `ffi_generateHmac` function extracts `payload` from incoming request using `JSON2STRUCT` macro, which will auto decode with base64.

References:
- https://github.com/truc0/ehsm/blob/274-move-the-hmac-verification-into-enclave/core/App/ffi_operation.cpp#L60
- https://github.com/truc0/ehsm/blob/274-move-the-hmac-verification-into-enclave/core/App/ffi_operation.cpp#L1573

ehsm_kms_service/function.js
truc0: Currently the original implementation `gen_hmac()` function is removed, while `gen_hmac_safe()` function has been renamed to `gen_hmac()`.
Since the original implementation of `gen_hmac()` and current implementation (same as `gen_hmac_safe()`) shares the same logic and have same output for certain input, the rename (`gen_hmac_safe()` -> `gen_hmac()`) should be safe.

yang8621 (approved):

syan10 (approved):

PR #298, opened 2023-08-28
check keyusage for symmetric key

whh39:
For symmetric key, keyusage must be EH_KEYUSAGE_ENCRYPT_DECRYPT

### DIFF
--- a/core/Enclave/enclave_hsm.cpp
+++ b/core/Enclave/enclave_hsm.cpp
@@ -438,7 +438,7 @@ sgx_status_t enclave_sign(const ehsm_keyblob_t *cmk, size_t cmk_size,
                           const ehsm_data_t *message, size_t message_size,
                           ehsm_data_t *signature, size_t signature_size)
 {
-    // TODO : make default padding mode for ECC/SM2 
+    // TODO : make default padding mode for ECC/SM2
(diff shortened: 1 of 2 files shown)


core/Enclave/enclave_hsm.cpp
```diff
@@ -169,6 +169,15 @@ sgx_status_t enclave_create_key(ehsm_keyblob_t *cmk, size_t cmk_size)
         return SGX_ERROR_INVALID_PARAMETER;
     }
 
+    if ((cmk->metadata.keyusage == EH_KEYUSAGE_SIGN_VERIFY && cmk->metadata.keyspec == EH_AES_GCM_128) ||
+        (cmk->metadata.keyusage == EH_KEYUSAGE_SIGN_VERIFY && cmk->metadata.keyspec == EH_AES_GCM_192) ||
+        (cmk->metadata.keyusage == EH_KEYUSAGE_SIGN_VERIFY && cmk->metadata.keyspec == EH_AES_GCM_256) ||
```
syan10, line 174: why not put this check into the switch...case dispatcher function, e.g,
ehsm_create_aes_key(cmk);
ehsm_create_sm4_key(cmk);

whh39: ok, that's better

syan10 (approved):

PR #302, opened 2023-09-04
fix: aad=NULL causes enclave_decrypt crash

truc0:
## Describe the bug

Calling `enclave_decrypt` with `aad=NULL` will crash when `cmk->metadata.keyspec` is `EH_AES_GCM_{128,192,256}`.

```c++
enclave_decrypt(cmk, cmk_size, NULL, 0, ciphertext, ciphertext_size, plaintext, plaintext_size);
```

This is because the `cmk` use `EH_AES_GCM_192` as keyspec, which transfer the control flow to `ehsm_aes_gcm_decrypt()`. However, the `enclave_decrypt()` function allows null `aad` while the `ehsm_aes_gcm_decrypt()` function does not check the `aad` varible before the following usage:

```cpp
// https://github.com/intel/ehsm/blob/main/core/Enclave/key_operation.cpp#L390
    ret = aes_gcm_decrypt(key,
                          plaintext->data,
                          block_mode,
                          cipherblob->data,
                          cipherblob->datalen - EH_AES_GCM_IV_SIZE - EH_AES_GCM_MAC_SIZE,
                          aad->data,  // BUG: DEREFERENCE OF NULL
                          aad->datalen, // BUG: DEREFERENCE OF NULL
                          iv,
                          SGX_AESGCM_IV_SIZE,
                          l_tag,
                          SGX_AESGCM_MAC_SIZE);
``` 

Then the enclave is crashed.

## Possible fix

A simple but work fix is changing the call to `aes_gcm_decrypt` in `ehsm_aes_gcm_decrypt` to the following:

```cpp
    ret = aes_gcm_decrypt(key,
                          plaintext->data,
                          block_mode,
                          cipherblob->data,
                          cipherblob->datalen - EH_AES_GCM_IV_SIZE - EH_AES_GCM_MAC_SIZE,
                          aad ? aad->data : NULL,
                          aad ? aad->datalen : NULL,
                          iv,
                          SGX_AESGCM_IV_SIZE,
                          l_tag,
                          SGX_AESGCM_MAC_SIZE);
```

Tests passed after the modification above.


core/Enclave/key_operation.cpp
```diff
@@ -387,8 +387,8 @@ sgx_status_t ehsm_aes_gcm_decrypt(ehsm_data_t *aad,
                           block_mode,
                           cipherblob->data,
                           cipherblob->datalen - EH_AES_GCM_IV_SIZE - EH_AES_GCM_MAC_SIZE,
-                          aad->data,
-                          aad->datalen,
+                          aad ? aad->data : NULL,
+                          aad ? aad->datalen : 0,
                           iv,
```
yang8621, line 392: https://github.com/intel/ehsm/blob/main/core/Enclave/key_operation.cpp#L199
Is there any similar bug in aes_gcm_encrypt() ?

truc0: `aes_gcm_encrypt()` only accesses `aad` when `aad` is not null.

Reference: https://github.com/intel/ehsm/blob/main/core/Enclave/openssl_operation.cpp#L94

yang8621: Sorry. I am talking about ehsm_aes_gcm_encrypt().  I mean if aad is null, will this [line ](https://github.com/intel/ehsm/blob/main/core/Enclave/key_operation.cpp#L199) in ehsm_aes_gcm_encrypt() also cause to null dereference?

truc0: Yes, `ehsm_aes_gcm_encrypt` has the same problem. Fixed in latest commit.

yang8621 (approved):

PR #306, opened 2023-09-14
Update the macro of MAX_ LENGTH to 6K.     

ForLisiteng:
Set the maximum length to 6K. However, since the input is a base64-encoded string, we need to adjust the checker accordingly.

### DIFF
--- a/ehsm_kms_service/payload_checker.js
+++ b/ehsm_kms_service/payload_checker.js
@@ -1,4 +1,4 @@
-const MAX_LENGTH = 8192
+const MAX_LENGTH = 6 * 1024
 const { ehsm_keySpec_t, ehsm_keyorigin_t, ehsm_message_type_t, ehsm_digest_mode_t, ehsm_padding_mode_t } = require('./constant')
 const { KMS_ACTION } = require('./apis')
 const { _result } = require('./function')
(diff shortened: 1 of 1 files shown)


yang8621 (approved):

ehsm_kms_service/payload_checker.js
```diff
@@ -1,4 +1,4 @@
-const MAX_LENGTH = 8192
+const MAX_LENGTH = 6144
```
syan10, line 1: change to 6*1024, to make it more clear

ehsm_kms_service/payload_checker.js
```diff
@@ -49,7 +49,7 @@ const keyid = {
 }
 const aad = {
   type: PARAM_DATA_TYPE.BASE64,
-  maxLength: MAX_LENGTH,
+  maxLength: MAX_LENGTH*4/3,
```
syan10, line 52: does it enough? and also please add comments to explain it's the real size after base64 encoding

ehsm_kms_service/payload_checker.js
```diff
@@ -81,7 +81,7 @@ const cryptographic_params = {
     keyid,
     ciphertext: {
       type: PARAM_DATA_TYPE.BASE64,
-      maxLength: MAX_LENGTH,
+      maxLength: (MAX_LENGTH+30)*4/3,
```
syan10, line 84: why +30？  iv+MAC? how about to align it to 32 bytes?

yang8621: It represents: b64 len: 4 * ceil(input_length / 3)

syan10 (approved):

PR #316, opened 2023-10-20
Implementing Python SDK

truc0:
# Introducing new Python SDK

## Motivation

Currently there is a CLI implementation of eHSM API client under `ehsm/test` directory. The current implementation provides the ability to interact with eHSM KMS server using CLI or Python code, but there are still some shortcomings:

1. Lack of type annotation: type annotation can improve developer experience and help writing high quality code
2. Different interfaces for each part of services: the return value of functions does not have a unified structure
3. Duplicate codes: some of the code is duplicated and can be improved

The new Python SDK plan to maintain the useful features (such as server tests, CLI) and addresses the shortcomings described above.

## Feature List and Roadmap

The new python SDK is divided into 3 parts: API client, CLI and server_test. Additional to these modules, the `serializers` module is added in order to provide type annotation and unified return structure.

### API Client

The API Client module encapsulates the communications to eHSM KMS server to functions and provides an unified entry, the features list is as following:

- [x] Implement cryptography API
- [x] Implement key management API
- [x] Implement remote attestation API
- [x] Implement secret management API
- [x] Handling failed request

### Serializers

Serializers module provides type of the return value of functions of API client. The features list is as following:

- [x] Implement response structure of cryptography API
- [x] Implement response structure of key management API
- [x] Implement response structure of remote attestation API
- [x] Implement response structure of secret management API

### Command Line Interface

The CLI module encapsulates the function call to API client and convert the arguments of function to CLI options. The features list is as following:

- [x] Implement CLI for cryptography API
- [x] Implement CLI of key management API
- [x] Implement CLI of remote attestation API
- [x] Implement CLI of secret management API
- [x] Call server test from CLI

### Server Tests

The server tests module includes tests of requesting eHSM KMS server by API client with different parameters, this module can be used to verify if the server has the same behavior when making changes to eHSM KMS server module, and also verify if the implementation of API client is correct.

The features list is as following:

- [x] Tests for cryptography API
- [x] Tests for key management API
- [x] Tests for remote attestation API
- [x] Tests for secret management API

### Utilities

- [x] Configuration files for releasing in PYPI (`pyproject.toml`, `requirements.txt` etc.)
- [x] Documentation (`README.md`, API description)
- [x] Squash commits


truc0:
## Usage of API client

Basic usage of the API client is shown as following:

```python
from ehsm.api import Client

client = Client()
client.enroll()

resp = client.create_key(keyspec=keyspec, origin=origin, keyusage=keyusage)
```

The raw response has the following content:

```json
{
  "code": 200,
  "message": "success",
  "result": {
    "keyid": "keyid123456"
  }
}
```

There are three styles for the return value (`resp` in the example above) API client function call.

### Style 1: `resp` is the `keyid`

This is the most intuitive response of this API. However, these return format will not be unified among the APIs since some response to the API has lots of variables in response.

Moreover, the user cannot get any information about the response object since the return value only contains `keyid`.

### Style 2: `resp` is the `result` and the raw response will be stored in `response` attribute

This style provides an unified format of all API client function calls. The user can also access the raw response object for more information if needed.

### Style 3: `resp` is an object with the same structure of the response content

This is the style of current implementation. It provides all information of the response object and keep the structure. This may be intuitive for an API client but it is not easy to use. User uses data in `result` mostly while `code` and `message` are not necessary most of the time. The user will have to append `.result` to the function call manually, which may not be elegant.

I'm looking for some advices in the style of the return value, feel free to comment.

truc0:
## Test for remote attestation

The server tests for remote attestation API is not implemented yet since:
1. This test requires external binary (`sgx_sign`), which means the test will fail when the user does not have SGX SDK installed
2. Currently the remote attestation API does not work on Docker environment, tests will fail if we think it is ok to fail when deploying in Docker

I'm still investigating why `generateQuote` and `verifyQuote` crashes inside Docker.

sdk/python/.gitignore
```diff
@@ -0,0 +1,162 @@
+.pdm-python
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
```
syan10, line 8: 这个里面都是需要的吗？ 感觉好像很多是你机器特有的？ 清理一下吧

truc0: This `.gitignore` file is copied from https://github.com/github/gitignore/blob/main/Python.gitignore and should cover most of the cache/intermediate files produced when running python

PR #324, opened 2023-11-22
fix: logs and DK are put on a tmpfs based folder

truc0:
Previously the log files and the assets (such as domain key) are located at `/var/run/ehsm`, which is dangerous since `/var/run` folder is based on tmpfs and will be erased after rebooting.

#322 fixed part of the problem by moving all of these files into `/var/local/ehsm`, but:
- the patch #322 only fixed the standalone mode, the mapping of `docker` and `k8s` still not changed
- moving all files (including runtime files like sockets) to `/var/local/ehsm` is simple but conflicts with the purpose of `/var/local` folder, some runtime files (e.g. sockets) should be placed in `/var/run/ehsm` folder

This PR fixes the two problems above by:
- move logs to `/var/log/ehsm`
- move DK and other assets to `/var/local/ehsm`
- keep socket files in `/var/run/ehsm`

All testcases are passed in standalone mode and in docker mode. The `k8s` mode is not tested.


buildenv.mk
```diff
@@ -42,7 +42,9 @@ EHSM_KMS_GIT_SHA=$(shell git rev-parse --short=7 --verify HEAD)
 ######## ehsm run folder  ########
 # the folder is used to store the files generated by the runtime
 # Please always keep the last /slash/
-RUNTIME_FOLDER := /var/local/ehsm/
+LOG_FOLDER := /var/log/ehsm/
+RUNTIME_FOLDER := /var/run/ehsm/
+LOCAL_DATA_FOLDER := /var/local/ehsm/
```
syan10, line 47: suggest to add a prefix for each macro, like:
EHSM_LOG_FOLDER
EHSM_RUNTIME_FOLDER

truc0: Thanks for suggesting. The changes are applied on the latest commit, please take a look.

yang8621 (approved):

docker/docker-compose.yml
```diff
@@ -38,6 +38,8 @@ services:
     volumes:
       - /var/run/aesmd:/var/run/aesmd
       - /var/run/ehsm:/var/run/ehsm
+      - /var/local/ehsm:/var/local/ehsm
```
syan10, line 41: can we use environments directly rather than hardcode?

truc0: Certainly, the latest commit implemented this, please check.

syan10 (approved):

